PR #1, opened 2021-12-21
[ENH] Adding setup code

adam2392:
Adds a package setup code:
- setup.py/cfg
- cleaner requirements files 
- a file directory structure for the WIP package, so we can start organizing functionality by files/folders

Adds unit tests in the same manner you already had them testing the consolidated feature functions w/ your implementation. In addition, tests them according to the Matlab input data.

I've discovered that the RPSD and Topoplot functions don't really work out of the box and haven't figured out why not yet. @jacobf18 for example, `testTopoPlot()` doesn't work in `test_feature_extraction.py`.


jacobf18 (approved):
Looks good

PR #2, opened 2022-01-21
Add questions via code

adam2392:
@jacobf18 Great start.

I took a look and ran things locally. Without a test file explicitly ran through EEGLab Matlab, hard to see what "differs". 

There's a few easy changes that I would like you to make, that will facilitate discussion on Monday:

- adding docstrings and some inline comments to `mne_to_eeglab_locs`. This function seems like it wouldn't change too much, so it'll be good to aggressively document what is going on and any description for obscure-like logic. You'll see I added "TODO" comments in various places.
- it'd be good to describe what `format_input()` is doing as well.
- as mentioned in slack, please run MATLAB version through for `eeglab_data_epochs_ica.set` sample data, saving the outputs of each step. We can then i) make our output match that of Matlab's as much as possible and ii) document where there are any diffs in a unit test. I can help you do this if you generate the data files.

### DIFF
--- a/ICLabel.py
+++ b/ICLabel.py
@@ -1,6 +1,8 @@
 import torch.nn as nn
 import torch
 import numpy as np
+import os
+import mne
 
 
 class ICLabelNetImg(nn.Module):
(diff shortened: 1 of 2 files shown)


ICLabel.py
```diff
@@ -173,35 +177,66 @@ def run_iclabel(images: np.ndarray, psds: np.ndarray, autocorr: np.ndarray) -> n
...
+    montage = raw.get_montage()
+    positions = montage.get_positions()
+    ch_pos = positions['ch_pos']
 
-    # %% Obtain carthesian coordinates
+    # get locations as a 2D array
+    locs = np.vstack(ch_pos.values())
+
+    # Obtain carthesian coordinates
     X = locs[:, 1]
```
adam2392, line 226: One question I had was why is X in position 1, but Y is in position 0?

Perhaps it's the orientation of MNE vs EEGLab? If so, it'll be useful to document what's going on, so another dev can follow.

jacobf18 (approved):

PR #18, opened 2022-04-22
MRG: Clean-up and add comparison/all implementations for feature extraction

mscheltienne:
Hey! @jacobf18 @adam2392 @anandsaini024 

Seeing all the good progress made, I wanted to finalize the first big step of this project: **a completely tested** and similar port of the network and features in Python. In this PR:

- I replaced ALL the test files with small tests files that were generated from the EEGLAB sample dataset. I also added the MATLAB scripts to generate those files and replicate all the results.
- Since we now know that the network has been perfectly ported, I focus on the feature extractions. I intend to go over all 3 and their implementations, in the coming couple of days (finally got some free time to work on this!).
- And a bit of clean-up / re-structuring

------------

Progress:
- [x] Topo
- [x] PSD
- [x] Autocorr

For each, I replicated as close as possible the MATLAB outputs and added the corresponding tests to compare the data arrays. Contrary to your code @jacobf18 that I used, I did not change the shapes and instead matched the shapes from MATLAB to simplify testing and comparison. 

And finally, the ICLabel feature extraction function differs depending on:
- raw or epoch
- If raw, how long? (3 behaviors for [0, 1] seconds for [1, 5] seconds and for above 5 seconds)

So I added test files for each scenario:
- `-epo` for epochs
- `-raw` for raws (above 5 secs)
- `-short-raw` for raws (1 to 5 secs)
- `-very-short-raw` for raws (0 to 1 sec)

--------------

Side note: 
- it is possible to match the random number generator between numpy and MATLAB in most cases. So we should be able to get good tests also for the PSD feature.
- trials represents epochs: 1 -> Raw; above 1 -> Epochs


mscheltienne:
Autocorr is done, it now matches at 100%.

mne_icalabel/tests/test_features.py
```diff
@@ -0,0 +1,354 @@
...
+        _eeg_rpsd_constants(epochs, ica)
+
+    # compute psdmed
+    psdmed = _eeg_rpsd_compute_psdmed(epochs, icaact, ncomp, nfreqs, n_points,
+                                      nyquist, index, window, subset_eeglab)
+
+    psdmed_eeglab = loadmat(psd_psdmed_epo_path)['psdmed']
+    assert np.allclose(psdmed, psdmed_eeglab, atol=5e-2)
+    # TODO: investigate why the tolerance had to be brought this high for this
+    # particular case.
```
mscheltienne, line 231: This one is weird. It's the first time I can't get by with a tolerance around `1e-4` (or even `1e-5)`.
The weirdest part is below, lines +297 to +302 where tolerance of `1e-5` is used for the same file _after_ formatting.

mne_icalabel/tests/test_features.py
```diff
@@ -0,0 +1,354 @@
...
+    ncomp, nfreqs, n_points, nyquist, index, window, _ = \
+        _eeg_rpsd_constants(epochs, ica)
+
+    # compute psdmed
+    psdmed = _eeg_rpsd_compute_psdmed(epochs, icaact, ncomp, nfreqs, n_points,
+                                      nyquist, index, window, subset_eeglab)
+
+    # format and compare
+    psd = _eeg_rpsd_format(psdmed)
+    assert np.allclose(psd, psd_eeglab, atol=1e-5)
```
mscheltienne, line 302: FYI previous comment. `psdmed` from this test function and `psdmed` from the previous test do match. After formatting, a tolerance of `1e-5` works, while this is not the case before formatting.

mscheltienne:
Test files have been re-created on the Windows on which I work for now, `sha1` checksum has been removed. I will recreate them on my macOS later.

mne_icalabel/tests/test_features.py
```diff
@@ -118,6 +123,30 @@ def test_compute_ica_activations():
     assert np.allclose(icaact, icaact_eeglab, atol=1e-4)
 
 
+# ----------------------------------------------------------------------------
+def test_loc():
+    """Test conversion of MNE montage to EEGLAB loc. This test works because
```
adam2392, line 128:
```suggestion
    """Test conversion of MNE montage to EEGLAB loc. 
    
    This test works because
```

mne_icalabel/tests/test_features.py
```diff
@@ -118,6 +123,30 @@ def test_compute_ica_activations():
...
+    MNE does the conversion form EEGLAB to MNE montage when loading the
```
adam2392, line 129:
```suggestion
    MNE does the conversion from EEGLAB to MNE montage when loading the
```

mne_icalabel/tests/test_features.py
```diff
@@ -118,6 +123,30 @@ def test_compute_ica_activations():
     assert np.allclose(icaact, icaact_eeglab, atol=1e-4)
 
 
+# ----------------------------------------------------------------------------
+def test_loc():
```
adam2392, line 127: I'm not sure I 100% follow the use of this test.

Is it to ensure that MNE doesn't break anything when loading in EEG coordinates stored from EEGLab?

mscheltienne: It's mostly a sanity check. If you load the `.set` in EEGLAB and in MNE, the electrode locations are read in the appropriate format for each. Thus, on the MNE side, if we read the `.set` file and convert the montage back to EEGLAB format with `mne_to_eeglab_locs`, the resulting `rd` and `th` have to match the location retrieved in EEGLAB.

mscheltienne:
I still have to look into the topoplot, probably tomorrow. I just checked the coordinates/montage (works as intended) and a quick test case with @jacobf18 function `ica_features.py:topoplot` that did not match.

mscheltienne:
Topo does match :)

I'll clean up tomorrow, fix feature formatting and add test for the entire pipeline.

mscheltienne:
All the individual steps and feature extraction works well and reproduce exactly the output from MATLAB.
But now, we enter into the 'weird' territory. The feature extraction starts by re-referencing to a common average, for both the data and ICA matrix.

`ICA_feature_extractor.m:L14:17`

```
% assuming chanlocs are correct
if ~strcmp(EEG.ref, 'averef')
    [~, EEG] = evalc('pop_reref(EEG, [], ''exclude'', setdiff(1:EEG.nbchan, EEG.icachansind));');
end
```

This step is problematic. If I save a dataset after re-referencing and reload it in EEGLAB.. the `icawinv` re-loaded is not the same! It looks like a bug in EEGLAB, ref. issue: https://github.com/sccn/eeglab/issues/462

I have to think a bit about this re-referencing and its impact. If you have any suggestion, please comment :)

mne_icalabel/tests/test_features.py
mscheltienne: solved.

mne_icalabel/tests/test_features.py
mscheltienne: solved itself, probably wrong file.

adam2392:
Whoot! Nice @mscheltienne if it matches, that's exciting news. This is a large diff, so I'll take some time soon to do an in-depth review.

@jacobf18 and @anandsaini024 should sanity check as well.

> I have to think a bit about this re-referencing and its impact. If you have any suggestion, please comment :)

Hmm so my opinion is that if it's happening in EEGLab, it's a "bug" but solely on their end. We simply want to match the EEGLab's first pipeline run (before saving the data to disc). Basically, I think it's fine.

A quick question, so if we re-reference to average multiple times it shouldn't change right? Perhaps misunderstanding, but why does `icawinv` change then?

mscheltienne:
For the `MANIFEST.in`, we can fix it in a follow-up. It needs more clean-up than just this `.txt`, and also do we include the data for the test files or not?

mne_icalabel/tests/test_features.py
```diff
@@ -0,0 +1,480 @@
...
+@pytest.mark.filterwarnings("ignore::RuntimeWarning")
+@pytest.mark.filterwarnings("ignore::FutureWarning")
+@pytest.mark.parametrize(
+    "file, psd_constant_file, eeglab_feature_file",
+    [
+        (raw_eeglab_path, psd_constants_raw_path, features_raw_path),
+        (epo_eeglab_path, psd_constants_epo_path, features_epo_path),
+    ],
+)
+def test_get_features(file, psd_constant_file, eeglab_feature_file):
```
adam2392, line 124:
```suggestion
def test_get_features_from_precomputed_ica(file, psd_constant_file, eeglab_feature_file):
```

Just to remember what the explicit purpose of the test is. Is this right?

mne_icalabel/tests/test_features.py
```diff
@@ -0,0 +1,480 @@
...
+    assert window_eeglab.shape[0] == 1
+    assert np.allclose(window, window_eeglab[0, :])
+
+    # for subsets, compare if the same elements are in both
+    assert subset_eeglab.shape[0] == 1
+    assert len(set(list(subset)).difference(set(list(subset_eeglab[0, :] - 1)))) == 0
+    assert len(set(list(subset_eeglab[0, :] - 1)).difference(set(list(subset)))) == 0
+
+
+def test_eeg_rpsd_compute_psdmed():
```
adam2392, line 310: It seems we don't need this extra test given the below `test_eeg_rpsd`(?)

mne_icalabel/tests/test_features.py
```diff
@@ -0,0 +1,480 @@
...
+        epochs, icaact, ncomp, nfreqs, n_points, nyquist, index, window, subset_eeglab
+    )
+
+    # format and compare
+    psd = _eeg_rpsd_format(psdmed)
+    assert np.allclose(psd, psd_eeglab, atol=1e-5)
+
+
+# ----------------------------------------------------------------------------
+def test_next_power_of_2():
```
adam2392, line 431: Let's move this to `test_utils.py`.

mne_icalabel/features.py
```diff
@@ -0,0 +1,479 @@
...
+            )
+
+    # normalize
+    psd = np.divide(psd.T, np.max(np.abs(psd), axis=-1)).T
+    # reshape and cast
+    return psd[:, :, np.newaxis, np.newaxis].transpose([2, 1, 3, 0]).astype(np.float32)
+
+
+# ----------------------------------------------------------------------------
+def _next_power_of_2(x) -> int:
```
adam2392, line 330: Let's move this to `utils.py`

adam2392 (approved):
Awesome refactor and sprint @mscheltienne !

It looks great to me. Just a few minor changes to maybe make things easier to maintain down the line.

mne_icalabel/tests/test_features.py
mscheltienne: Fine by me, the only added information was that if only `test_eeg_rpsd` fails, then the problem lies within `_eeg_rpsd_format`. But that is not worth an additional test and can be debugged on the fly if `test_eeg_rpsd` ever fails.

mscheltienne:
I am getting an error with `make black` following the last commit. I am not familiar with it and the traceback is.. lacking. Any idea?

adam2392:
> I am getting an error with `make black` following the last commit. I am not familiar with it and the traceback is.. lacking. Any idea?

Sorry the make recipe only "checks" for formatting errors. 

You'll need to run "black ."

I think I can fix that in another PR to make the dev exp easier.

adam2392:
Sorry... just noticed that. I've previously set it up to just format it.

adam2392:
@jacobf18 and @anandsaini024 skim the changes to see if you spot any issues. Meanwhile ill merge this.

Issue #19, opened 2022-04-28
Discussion about the API for mne_icalabel and mne


mscheltienne:
I would like to discuss a bit what the API would look like, both on the `mne_icalabel` side and on the `mne` side. Here is my point of view:

### `mne_icalabel`

As described in the readme, this package is not only a port of the popular `ICLabel` network for EEG but could host other models for EEG, MEG, and iEEG. Thus, I would structure the package as:

```
mne_icalabel
└─ iclabel
    └─ assets
    └─ tests
    └─ network.py
    └─ features.py
└─ model2
└─ model3
└─ ...
└─ icalabel.py
```

The main entry point from a user perspective would be in `mne_icalabel/icalabel.py` through a function that takes as input a `raw` or `epochs` instance, an ICA decomposition, and a method (str). The `method` argument would have a couple of accepted values: the name of the models, e.g. `'iclabel'`; and would be responsible for selecting the model used.
Based on the model used, the number of class in which the components are classified may vary. But in any case, the output should be the label (which class) and the probability that each component is in each class (scores). 

Additionally to this main entry point, I would keep public a function to retrieve the features of a model, and a function to run a set of features through the model. e.g. for ICLabel, it would be `mne_icalabel/iclabel/features.py:get_features` and `mne_icalabel/iclabel/network.py:run_iclabel`. Everything else would become private. 

### `mne`

In version 1.1, there are already 4 methods to find bad components in `mne.preprocessing.ICA`: `find_bads_ecg`, `find_bads_eog`, `find_bads_muscle` and `find_bads_ref`. Following this pattern, I would go for a `find_bads` method that would wrap around the entry point in `mne_icalabel/icalabel.py`, with the 2 parameters `inst` and `method`. As for the 4 other methods, the label can be stored in `ica.labels_`. 

I think it would be worth also storing the probability that each component is in each class (scores) in a similar attribute to `ica.labels_`, for both this new `find_bads` method and for all others as well. 
I brought the idea here a while ago: https://github.com/mne-tools/mne-python/issues/9846 but I didn't follow up on it yet. It does require a new ICA FIFF field for a 2D matrix of floats. 

--------

Finally, from an MNE perspective, the API could be simplified a bit further. At term, the `find_bads` method could be the only one remaining with the 4 others moved to `mne_icalabel` as models and called through `find_bads` with the correct method keyword specified.

adam2392:
> ### `mne_icalabel`
[...]

PR #21, opened 2022-04-28
MRG: Update repository structure

mscheltienne:
As per #19, the structure could be something like that.

I chose `label` as name for the entry-point: adding one for the entire package (`mne_icalabel`) and one for each model (submodule, only `iclabel` for now). If you have a better name in mind, please comment 😉 
That would bring the public API to:

```
from mne_icalabel import label
label(raw, ica, method='iclabel')
```

And

```
from mne_icalabel.iclabel import label
label(raw, ica)
```

--------------------

I added an underscore to all private functions. I chose to keep the feature extraction `get_features`, the network `ICLabelNet`, the forward pass `run_iclabel` public on top of the entry-point `label`. 

WDYT?


adam2392:
Thanks for getting to this right away!

I feel like `label` might be too generic and not explicit enough. Perhaps `label_ic`?

mscheltienne:
How about `label_components` since MNE has the methods `get_components` and `plot_components` for ICA?

adam2392:
> How about `label_components` since MNE has the methods `get_components` and `plot_components` for ICA?

I like it!

mscheltienne:
Done, and I added the simplest of test cases to run `label_components` on the MNE raw sample dataset.. and it's already failing! 😅

mscheltienne:
For the error, it's because autocorrelation ends with a resampling: 

```
# resample to 1 second at 100 samples/sec
resamp = resample_poly(ac.T, 100, np.round(raw.info["sfreq"])).T
```

The sample dataset sampling rate is `600.614990234375` Hz. It was complaining that the up-scaling and the down-scaling factors must be integers, so I rounded `raw.info['sfreq']`. But that's not the correct fix as now, instead of resampling to 100 samples, it resampled to 99.. thus failing the reshaping at a later stage.

I am not sure how to correctly fix this at the moment, I'll think about it.

mscheltienne:
I agree that some error checking is necessary with some warning:
- Is the reference a common average? (we need more info than `'custom_ref_applied'` provide, there is an issue somewhere on the main repo about storing the reference for EEG dataset)
- Is the sampling frequency an integer? (not sure, but I think a float sfreq would crash in EEGLAB)
- Is the dataset BP between 0 and 100 Hz?
and some raises:
- Do we have the same channel in the ICA and in the instance? (ideally, you need to provide the instance used for fit)

We'll discuss those bullet points later, but for the frequency, I don't think we should force it to be integer because if a user has a dataset like the MNE sample, he can not easily change the sampling frequency. The field `.info['sfreq']` is among the lock fields that should not be tampered with and raises if you try to directly change it. Resampling doesn't seem like the way to go either as it will alter the data.. for no real reason.

---------------------

Instead, I would simply figure out if we have to floor or ceil the sampling frequency in the autocorrelation function to get the 100 samples. I am not (yet) convinced `floor` for every case is the correct method.

adam2392:
> I agree that some error checking is necessary with some warning:

Agreed w/ all your warning proposals. Shall we merge this and go for it in another PR?

> We'll discuss those bullet points later, but for the frequency, I don't think we should force it to be integer because if a user has a dataset like the MNE sample, he can not easily change the sampling frequency. The field `.info['sfreq']` is among the lock fields that should not be tampered with and raises if you try to directly change it. Resampling doesn't seem like the way to go either as it will alter the data.. for no real reason.

Perhaps we have an optional argument `sfreq=None`, which is used to override the sampling rate. I think intelligently flooring/ceiling it is not possible since one could easily have `127.9584` or `128.12040` for example sampling rate, which actually corresponds to 128 Hz. WDYT of this soln? Con is adding an additional kwarg, but I don't see an easy way around w/o too much hacking.

mne_icalabel/iclabel/tests/test_label_components.py
```diff
@@ -0,0 +1,25 @@
...
+# preprocess
+raw.filter(l_freq=1.0, h_freq=100.0)
+raw.set_eeg_reference("average")
+# fit ICA
+ica = ICA(n_components=15, method="picard")
+ica.fit(raw)
+
+
+@pytest.mark.filterwarnings("ignore::RuntimeWarning")
+@pytest.mark.filterwarnings("ignore::FutureWarning")
```
adam2392, line 21: I will clean this up in the next PR.

adam2392 (approved):

mne_icalabel/iclabel/features.py
```diff
@@ -7,7 +7,7 @@
 from numpy.typing import NDArray
 from scipy.signal import resample_poly
 
-from .utils import _next_power_of_2, gdatav4, mne_to_eeglab_locs, pol2cart
+from .utils import _gdatav4, _mne_to_eeglab_locs, _next_power_of_2, _pol2cart
 
 
 def get_features(inst: Union[BaseRaw, BaseEpochs], ica: ICA):
```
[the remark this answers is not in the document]
mscheltienne: Yep, good idea.

mne_icalabel/iclabel/tests/test_label_components.py
mscheltienne: Much appreciated!

mscheltienne:
For the warnings, yes I will add those later on. 
For the optional ICA, sounds good to me! But we will have to document which settings we use by default. 

For the sampling rate, I think intelligent flooring/ceiling is possible, but your example is a very good point and I suspect it will fail with the current code. I am not that convince by the additional argument which complicates needlessly the API.
I will add a test case for the resampling alone, and hopefully a solution that will convince both of us that correct flooring/ceiling is possible :)

adam2392:
Sounds good! FYI the repo is moved to my GitHub for now so I can work on the CI and docs before doing another migration to mne.tools :p 

You'll have to reset your remote URL

`git remote remove <current one>`
then
`git add remote <new one>`

mscheltienne:
Perfect, don't worry about the move(s) ;)

PR #26, opened 2022-04-29
Improve testing for resampling in the autocorrelation feature

mscheltienne:
It looks like I got it right in #21.
WDYT?

### DIFF
--- a/mne_icalabel/iclabel/features.py
+++ b/mne_icalabel/iclabel/features.py
@@ -387,12 +387,7 @@ def _eeg_autocorr_welch(raw: BaseRaw, ica: ICA, icaact: NDArray[float]) -> NDArr
     ac = np.divide(ac, den)
 
     # resample to 1 second at 100 samples/sec
-    # i.e. the resampling must output an array of shape (components, 101), thus
-    # respecting '100 < ac.T.shape[0] * 100 / down <= 101'.
-    down = int(raw.info["sfreq"])
-    if 101 < ac.shape[1] * 100 / down:
-        down += 1
-    resamp = resample_poly(ac.T, 100, down).T
+    resamp = _resample(ac, raw.info['sfreq'])
     resamp = resamp[:, 1:, np.newaxis, np.newaxis].transpose([2, 1, 3, 0])
     return np.real(resamp).astype(np.float32)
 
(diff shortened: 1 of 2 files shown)


adam2392:
Wouldn't this only handle sampling rates of 100 Hz? What if it was 200? 

Maybe we should try handling all common sampling rates of EEG hardware? So I'm guessing like 50, 100, 128, 200, 256? 

Idk, can get complicated though :/

mscheltienne:
Well, that's what EEGLAB does. The autocorrelation functions ends with:

```
% resample to 1 second at 100 samples/sec
resamp = resample(ac', 100, EEG.srate)';
```

Which corresponds to [`resample_poly`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.signal.resample_poly.html) with `up` hard coded to 100, and `down` set to `sfreq`. 

All it's doing in #21 and in this PR is supporting floating point `sfreq` by setting the `down` argument either to `int(sfreq)` or to `int(sfreq) + 1`, in order to get the correct output shape `(n_comp, 101)`.
In EEGLAB, since the `sfreq` is always an integer, this step is not needed because if `down=EEG.srate` (int), then the output shape is always `(n_comp, 101)`.

adam2392:
> Well, that's what EEGLAB does. The autocorrelation functions ends with:
> 
> ```
> % resample to 1 second at 100 samples/sec
> resamp = resample(ac', 100, EEG.srate)';
> ```
> 
> Which corresponds to [`resample_poly`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.signal.resample_poly.html) with `up` hard coded to 100, and `down` set to `sfreq`.
> 
> All it's doing in #21 and in this PR is supporting floating point `sfreq` by setting the `down` argument either to `int(sfreq)` or to `int(sfreq) + 1`, in order to get the correct output shape `(n_comp, 101)`. In EEGLAB, since the `sfreq` is always an integer, this step is not needed because if `down=EEG.srate` (int), then the output shape is always `(n_comp, 101)`.

Fair I guess that's what ICLabel does so at least for this version! Unfortunate, I wonder how sensitive ICLabel is to the sampling rate... Room for improvement :)

adam2392 (approved):

Issue #5, opened 2022-03-04
Discussion of correctness of IClabel-python vs EEGLab


adam2392:
@mscheltienne reposting here as an issue to begin discussion again.

Quick update: our intern should have started on the 1st of March. But he did not yet get his work permit (mandatory in Switzerland), thus his start has been delayed and we hope he will be on board next week. 

-------

I finally took the time to test it! It looks very promising but seems to favor a lot 'Other'.
Am I correct in assuming that the classes outputted are in the order:

```
classes = ['Brain', 'Muscle', 'Eye', 'Heart', 'Line Noise', 'Channel Noise', 'Other']
```

I ran it on [this file](https://drive.google.com/file/d/1oqoRqLpYJVg65x-KzCmq8ezkXXJmGFtU/view?usp=sharing) quickly:

```
from matplotlib import pyplot as plt
import mne
import numpy as np

from mne_icalabel.ica_label import ica_eeg_features
from mne_icalabel.ica_net import run_iclabel

#%% Load
raw = mne.io.read_raw_fif('ica-test-raw.fif', preload=True)

#%% ICA
ica = mne.preprocessing.ICA(method='picard', max_iter='auto')
ica.fit(raw, picks='eeg')

#%% Features
features = ica_eeg_features(raw, ica)
topo = features[0].astype(np.float32)
psds = features[1].astype(np.float32)
autocorr = features[2].astype(np.float32)
labels = run_iclabel(topo, psds, autocorr)

classes = ['Brain', 'Muscle', 'Eye', 'Heart', 'Line Noise', 'Channel Noise', 
           'Other']

#%% Plot
f, axes = plt.subplots(1, 7, sharex=True, sharey=True, figsize=(20, 5))
for k, ax in enumerate(axes):
    ax.bar(np.arange(0, labels.shape[0]), labels[:, k])
    ax.set_title(classes[k])

for i, label in enumerate(labels):
    k = np.argmax(label)
    axes[k].bar(i, label[k], color='crimson')
    
f.tight_layout()

#%% ICA plots
ica.plot_components()
ica.plot_sources(raw)
```

![Screenshot 2022-03-04 at 16 40 40](https://user-images.githubusercontent.com/73893616/156793846-a85b2103-1c6d-4e3b-88fd-38703dd737ae.png)

Eye and heartbeat are correct. The last time we talked, you told me it failed at classifying even simple blinks, did you find what caused this?

[...]

PR #31, opened 2022-05-01
[ENH] Moredocs and experiment with api implementation

adam2392:
PR Description
--------------

Addresses #5 and #19

Merge checklist
---------------

- [ ] All comments resolved
- [ ] This is not your own PR
- [ ] All CIs are happy

### DIFF
--- a/mne_icalabel/label_components.py
+++ b/mne_icalabel/label_components.py
@@ -1,10 +1,12 @@
 from typing import Union
 
+import numpy as np
 from mne import BaseEpochs
 from mne.io import BaseRaw
 from mne.preprocessing import ICA
 from mne.utils import _validate_type
 from mne.utils.check import _check_option
+from sklearn.base import TransformerMixin, BaseEstimator
 
 from .iclabel import label_components as label_components_iclabel
 from .utils import _validate_inst_and_ica
@@ -12,6 +14,26 @@ from .utils import _validate_inst_and_ica
 methods = {
     "iclabel": label_components_iclabel,
 }
+
+class AutoLabelICA(TransformerMixin):
+    def __init__(self, method:str ='iclabel') -> None:
+        self.method = method
+
+    def fit(self, X, y):
+        pass
+
+    def transform(self, raw, ica):
+        ic_labels = label_components(raw, ica, method=self.method)
+
+        # Afterwards, we can hard threshold the probability values to assign
+        # each component to be kept or not (i.e. it is part of brain signal).
+        # The first component was visually an artifact, which was captured
+        # for certain.
+        not_brain_index = np.argmax(ic_labels, axis=1) != 0
+        exclude_idx = np.argwhere(not_brain_index).squeeze()
+
+        ica.apply(raw, exclude=exclude_idx)
+        return raw
 
 
 def label_components(inst: Union[BaseRaw, BaseEpochs], ica: ICA, method: str):
--- a/mne_icalabel/label_components.py
+++ b/mne_icalabel/label_components.py
@@ -6,34 +6,14 @@ from mne.io import BaseRaw
 from mne.preprocessing import ICA
 from mne.utils import _validate_type
 from mne.utils.check import _check_option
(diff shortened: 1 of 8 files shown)


mscheltienne:
One think that could be useful is to keep the `np.ndarray (n_components, n_classes)` format which has merit in itself for `from mne_icalabel.iclabel import label_components` while providing a 'simpler' user-friendly output for `from mne_icalabel import label_components`.

mne_icalabel/label_components.py
```diff
@@ -30,10 +32,38 @@ def label_components(inst: Union[BaseRaw, BaseEpochs], ica: ICA, method: str):
...
+    labels_pred_proba = methods[method](inst, ica)
+    labels_pred = np.argmax(labels_pred_proba, axis=1)
+    labels = [ICLABEL_NUMERICAL_TO_STRING[label] for label in labels_pred]
+
+    component_dict = {
+        "y_pred_proba": labels_pred_proba,
+        "y_pred": labels_pred,
+        "labels": labels,
+    }
+    return component_dict
```
mscheltienne, line 69: Any reason for nesting the output in a dictionary? Can't we directly have `return y_pred_proba, y_pred, labels` (in this order or another)?

mscheltienne (approved):
One comment, otherwise LGTM

mne_icalabel/label_components.py
adam2392: No real preference.

I only did this to make it super explicit and limit the number of return variables / make it a one-liner to convert to a Dataframe. Would you prefer the 3 tuple return instead?

mscheltienne: No strong preference either, but you can't convert this to a DataFrame that easily either since `labels_pred_proba` is a 2D array.

adam2392: Ah fair. Hmm now that I think about it more I do prefer in general single return arguments, but I'm also unsure if this is optimal here... I'll leave this open for now and we can always change it.

adam2392:
Nice, I also added rendering docs directly with a circleCI redirect: 

https://output.circle-artifacts.com/output/job/af197406-a7a9-45c9-8093-6c3c0d5c2043/artifacts/0/dev/auto_examples/iclabel_automatic_artifact_correction_ica.html#sphx-glr-auto-examples-iclabel-automatic-artifact-correction-ica-py

adam2392:
I will merge this once the CI complete.

mne_icalabel/label_components.py
mscheltienne: I like the idea of simple conversion to DataFrame for the user (of course not on our end, let's not have pandas as a dependency just for that). With that in mind, I see 2 individual DataFrame in the proposed output:

| idx | labels | y_pred |
|-----|-------|--------|
| 0   | Eye   | 0.89   |
| 1   | Brain | 0.67   |
| ... | ...   | ...    |

and:

| idx    | Brain | Muscle | Eye  | Heart | Line Noise | Channel Noise | Other |
|--------|-------|--------|------|-------|------------|---------------|-------|
| 0      | 0.2   | 0.89   | 0.1  | 0.05  | 0.03       | 0.07          | 0.08  |
| 1      | 0.03  | 0.67   | 0.06 | 0.04  | 0.1        | 0.3           | 0.09  |
| ...    | ...   | ...    | ...  | ...   | ...        | ...           | ...   |
| n_comp | 0.3   | 0.23   | 0.43 | 0.22  | 0.11       | 0.08          | 0.7   |

So how about 2 outputs: `return dict(labels=labels, y_pred=labels_pred), {value: labels_pred_proba[:, k] for k, value in ICLABEL_NUMERICAL_TO_STRING.items()}`.

IMO, for what needs to be stored in the mne ICA instance, we do not need the `y_pred`, but only the `labels` and the second dict.

With this solution, you can get the 'raw' numpy array of probabilities with `mne_icalabel.iclabel.iclabel_label_compoments` or dictionaries ready to be transformed to DataFrames with  `mne_icalabel.label_components`.

WDYT?

PR #40, opened 2022-05-08
Add default ICA decomposition and default method

mscheltienne:
As discussed, we could fit a sensible ICA if None is provided. This PR is a draft where I chose to use `picard`. I also added a default to the `method` argument in `label_components`. The idea is to be able to do:

```
from mne_icalabel import label_components

labels, ica = label_components(raw)
```

Personally, I'm **-1** for this change. I prefer to leave the responsibility of getting an ICA (and hopefully a good one) in the hand of the user. Adding the possibility to 'auto-fit' an ICA makes `label_components` messier, since it now has to also output the ICA decomposition to which the labels relate to.

@adam2392 WDYT?


jacobf18:
I am also in favor of leaving the ICA to the user. However, I think we should be clear that EEGLAB uses extended infomax for ICA. I have not tested the performance of ICLabel on different types of ICA methods, but I think it should be able to work on any type

mscheltienne:
Good point! It should, but we should probably issue a warning if the ICA decomposition provided does not use an extended infomax.

hoechenberger:
Note that picard can produce the same results as extended infomax if the following parameters are passed:

```python
ortho=False, extended=True
```

adam2392:
+1 for keeping current workflow and making users run ICA themselves

Issue #48, opened 2022-05-10
Remove extras_require?


hoechenberger:
All dependencies currently listed as `extras_require.full` apparently are only used for building docs and running tests, so I think they should be dropped from `setup.cfg`.

Issue #47, opened 2022-05-10
Is MNE 0.24 supported?


hoechenberger:
setup.cfg says mne 0.24 is the minimum required version. However, I don't think this has (ever?) been tested? Should this be bumped to 1.0?

mscheltienne:
+1 to bump to 1.0.

PR #49, opened 2022-05-10
Fix requirements and python version

### LINKED ISSUE #46: Mismatch between required Python version according to docs vs setup.cfg
In setup.cfg, it says Python 3.7+ is required, but in the docs, it says 3.8+

mscheltienne:
Closes: #46
Closes: #47
Closes: #48
Closes: #52

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,0 @@
-numpy
-scipy
-mne
-torch
-importlib-resources; python_version<"3.9"
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,3 +2,4 @@ numpy
 scipy
 mne
 torch
+importlib-resources; python_version<"3.9"
--- a/setup.cfg
+++ b/setup.cfg
@@ -80,8 +80,7 @@ match_dir = ^(?!\.|doc|examples).*$
 add-ignore = D100,D104,D107,D413
 add-select = D214,D215,D404,D405,D406,D407,D408,D409,D410,D411
 ignore-decorators = property
-match =
-    (?!_eeg_features).*\.py
+match = (?!_eeg_features).*\.py
 
 [coverage:run]
 omit =
--- a/setup.cfg
+++ b/setup.cfg
@@ -39,6 +39,7 @@ install_requires =
   scipy >= 1.2.0
   torch
   setuptools
+  importlib-resources; python_version<"3.9"
 packages = find:
 include_package_data = True
 
--- a/setup.cfg
+++ b/setup.cfg
@@ -26,6 +26,7 @@ classifiers =
   Programming Language :: Python :: 3.7
   Programming Language :: Python :: 3.8
   Programming Language :: Python :: 3.9
+  Programming Language :: Python :: 3.10
 project_urls =
   Documentation = https://mne.tools/mne-icalabel
   Bug Reports = https://github.com/mne-tools/mne-icalabel/issues
--- a/setup.cfg
+++ b/setup.cfg
@@ -81,8 +81,7 @@ match_dir = ^(?!\.|doc|examples).*$
 add-ignore = D100,D104,D107,D413
 add-select = D214,D215,D404,D405,D406,D407,D408,D409,D410,D411
 ignore-decorators = property
-match =
-    (?!_eeg_features).*\.py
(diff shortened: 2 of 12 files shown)


hoechenberger:
> Fix #46, #47 and #48.

btw this will only close #46 and reference the others. You need to explicitly "fixes" or "closes" every single one :(

mscheltienne:
I know.. GitHub isn't yet that smart, and yet I keep trying 😅 I'll close manually for this one.

hoechenberger:
> I know.. GitHub isn't yet that smart, and yet I keep trying 😅 I'll close manually for this one.

In fact I believe this **used** to work, and at one point it didn't anymore :(

hoechenberger:
This also once again shows the dangers of using `requirements.txt` … just saying :)

setup.cfg
```diff
@@ -87,7 +80,7 @@ match_dir = ^(?!\.|doc|examples).*$
 add-ignore = D100,D104,D107,D413
 add-select = D214,D215,D404,D405,D406,D407,D408,D409,D410,D411
 ignore-decorators = property
-match = 
+match =
     (?!_eeg_features).*\.py
```
mscheltienne, line 84: @adam2392 This looks like an old match for one of the files that were in the `tests` folder. Also, if this is not on a separate line but on the same line -> `match = (?!_eeg_features).*\.py`, `pydocstyle` complains a lot about our current docstrings. 
I guess this match should be updated with an exclusion for the `tests` .py files?

adam2392: Ah good catch... then the pydocstyle was passing all along mistakenly? :(

requirements.txt
```diff
@@ -1,4 +1,5 @@
 numpy
 scipy
 mne
-torch
\ No newline at end of file
+torch
+importlib-resources; python_version<"3.9"
```
hoechenberger, line 5: I don't think it makes much sense listing this here unless you pin it to a specific version, as it's listed as a package requirement in setup.cfg already

mscheltienne: If we want to keep `requirements.txt`, then it should have the same information as `setup.cfg`. The goal is to enable `pip install -r requirements.txt`. Now, personally, I don't like duplicating code/info across files.
So either we read the `requirements.txt` in `setup.py` and remove it from `setup.cfg` (as MNE) or we remove the `req` TXT files entirely.

I already know which option you would choose @hoechenberger ;)
@adam2392 Any preference?

adam2392: I always install things via `requirements.txt` file when doing dev work... so I feel like that's nice to have(?) 

@hoechenberger is your proposal to have both?

hoechenberger: But I suppose it's a longer discussion. Feel free to merge in either case ;)

adam2392: I see... I guess I'm in favor of `setup.cfg` and then extra requirements files for devs(?)

hoechenberger: btw this is what the official Python docs say:

https://packaging.python.org/en/latest/discussions/install-requires-vs-requirements/#requirements-files

And this Medium post also gives an opinion:
https://towardsdatascience.com/requirements-vs-setuptools-python-ae3ee66e28af#bb0e

adam2392: So we still need the req files for doc and testing since those are useful but based on skimming these two we should get rid of requirements.txt file in favor of setup.cfg

Sounds good to me! I like simplicity

hoechenberger: No you could just move those to extras_require too

dev
docs

Something like that :)

But it's opinionated! So in the end it comes down to your personal preferences

mscheltienne: Personally, I also appreciate the `req.txt` files for dev. 
But I do think it would be nice if we can avoid duplicating the req. between `setup.cfg` and the `.txt` files. Removing the base `requirements.txt` in favor of `setup.cfg` while preserving the doc and test `.txt` works for me.

hoechenberger: Sounds like a good compromise for now!

mscheltienne:
@adam2392 I updated a bit the CIs, please have a look and merge if happy.

adam2392 (approved):

adam2392:
Thanks! @mscheltienne

PR #57, opened 2022-05-11
[DOC] Try circleCI

adam2392:
Thanks for contributing. If this is your first time,
make sure to read the [contributing guide.](https://github.com/mne-tools/mne-python/blob/main/CONTRIBUTING.md)

PR Description
--------------

Describe your PR here

Merge checklist
---------------

- [ ] All comments resolved
- [ ] This is not your own PR
- [ ] All CIs are happy

### DIFF
--- a/examples/iclabel_automatic_artifact_correction_ica.py
+++ b/examples/iclabel_automatic_artifact_correction_ica.py
@@ -164,7 +164,7 @@ ica.plot_sources(raw, show_scrollbars=False)
 # .. _`this EEGLAB tutorial`: https://labeling.ucsd.edu/tutorial/labels
 
 
-# sphinx_gallery_thumbnail_number = 9
+# sphinx_gallery_thumbnail_number = 1
 ica.plot_components()
 
 # blinks
--- a/setup.cfg
+++ b/setup.cfg
@@ -6,7 +6,7 @@ author = mne-icalabel developers
 maintainer = Adam Li, Jacob Feitelberg, Patrick Myers
 maintainer_email = ali39@jhu.edu
 description = MNE-ICALabel: Automatic labeling of ICA components from MEG, EEG and iEEG data with MNE.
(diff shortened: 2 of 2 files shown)


mscheltienne:
I am a bit surprised the fastICA did not converge as I've increased the tolerance and fixed the random seed to a value that did not fail the CIs on the other PR or on the push to main.. maybe you can increase the tolerance further, to e.g. 1e-1 for this mock ICA.

agramfort:
FastICA can fail to converge !

« The orthogonal approach is followed by many ICA algorithms. Among them, FastICA [2] stands out by its simplicity, good scaling and a built-in ability to deal with both sub-Gaussian and super-Gaussian components. It also enjoys an impressive convergence speed when applied to data which are actual mixture of independent components [3, 4]. However, real data can rarely be accurately modeled as mixtures of independent components. In that case, the convergence of FastICA may be impaired or even not happen at all. »

Taken from:

https://hal.inria.fr/hal-01651842/document

mscheltienne:
Thank you! Applying ICA on this mock data `np.random.randint(low=1, high=10, size=(6, 10000)) / 1000` isn't working well 😅 
But if the `random_state` is set, shouldn't fastICA always output the same value, and thus always converge if that seed leads to convergence?

Anyway, I'll update the test, with a fake or a loaded ICA instance, since we don't actually need to fit valid ICAs for this test.

hoechenberger:
The RNG can behave differently in different systems, so it might work locally with a certain seed, but produce different results on CI with the exact same seed; at least this is what I believe I've seen in the past.

mscheltienne:
That I could understand, but it passed the same CIs twice with this seed.

hoechenberger:
But especially in CI you might end up in different machines for different runs! Maybe that's part of the issue

mscheltienne:
#58 should solve it.

PR #65, opened 2022-06-14
Remove unused azure pipeline

mscheltienne:
@adam2392 This repo does not use azure, right?


adam2392:
We should! For windows testing.

larsoner:
You can use Azure or GH actions for windows. GH Actions uses Azure under the hood IIRC anyway. So if you're happy with GH actions you can just go ahead and use it

adam2392:
Okay! @mscheltienne let's just get rid of it then.

Seems like the unit tests failed once the mne-testing-data was updated. Perhaps it's using a cached version of the data? Should we just re-run the CI?

larsoner:
No, downloading of the mne-testing-data releases is totally broken at the GH end. You'll need to test locally. Feel free to upvote this, maybe someone will notice

https://github.com/orgs/github-community/discussions/18543

Issue #12, opened 2022-04-06
API and GUI: Creating a highly informative example/jupyter notebook for IC labeling


adam2392:
We'll need an example/notebook walking through any team member on how the guidelines for labeling IC components are.

Ideally this notebook/example can be used by team asynchronously to 
1) load the data 
2) run ICA 
3) manually annotate the components with their best guess and then 
4) save the labels and ICA to centralized repo.

Reference: https://www.sciencedirect.com/science/article/pii/S0165027015000928?via%3Dihub

mscheltienne:
Additional reference:
- https://labeling.ucsd.edu/tutorial/labels
- https://labeling.ucsd.edu/labelfeedback

adam2392:
^ In addition to the above tutorials, and in line with discussion today on Zoom, we want a lightweight pipeline in-house to annotate IC components that has the following features:

- GUI to easily have user visualize IC components from BIDS formatted dataset stored on disc
- interface to select different IC components as different labels (e.g. brain, eye blink, heart beat, etc.)
- 1-click commands to save to disc, preferably in line with BIDS format

I wonder if others have already done so. I can ask in the MNE-core dev meeting.

@mscheltienne brought up mnelab, which might have a subset of the above features.

adam2392:
For discussion next week: I was thinking that if we want to produce an "annotation" pipeline, we want to fire up the following subplots:

- topoplot
- the actual IC time-series
- power spectrum of the IC time-series
- dipole plot (is this easy to compute?)
- single-trial time-course (?)

Possibly also with a corresponding time-series plot of all ICA components. This subplot would be repeated for all ICA components, and the user would be required to create a list of ICA component labels. The question is how to make this interactive. At the very minimum, we should have a "utility" function perhaps in mne-icalabel that automatically produces all those subplots for a single ICA component, perhaps something like:

`plot_ica_features(raw, ica, component_index=1)`

Following the plots shown here: https://labeling.ucsd.edu/tutorial/labels

mscheltienne:
Agree. The single-trial time-course is optional I think, personally, I'm mostly looking at the 3 first features you listed.
The utility function is a good idea, and it would be nice if it was easy to embed the figure(s) into a Qt widget. 

Also, we need to decide on the label classes:
-> Brain / No brain (as I think is the case on swap4ica and on the MARA dataset)
-> Brain / Ocular / Cardiac / Muscular / ... (as e.g. ICLabel)

[...]

PR #66, opened 2022-06-15
[ENH] Draft GUI for labeling components

adam2392:
PR Description
--------------

Addresses: #12 

From MNE-Python, I understand there are two parts of a GUI API pattern that we can follow:

- actual GUI Python class: https://github.com/mne-tools/mne-python/blob/04b5ed9afb7722e72e43dd2082bdfc5e4c1c9d52/mne/gui/_ieeg_locate_gui.py#L106-L122
- functional API: https://github.com/mne-tools/mne-python/blob/04b5ed9afb7722e72e43dd2082bdfc5e4c1c9d52/mne/gui/__init__.py#L191-L229

I'm not very familiar with a good workflow for getting this to work properly so can definitely use help :p 

### Details of the GUI

The GUI will I think house 3 plots, stemming from `ICA.plot_properties()` (https://mne.tools/stable/generated/mne.preprocessing.ICA.html#mne.preprocessing.ICA.plot_properties). There should be basic saving/loading functions of course that interface with the BIDS files on disc.

Merge checklist
---------------

- [ ] All comments resolved
- [ ] This is not your own PR
- [ ] All CIs are happy


mne_icalabel/gui/__init__.py
```diff
@@ -0,0 +1,31 @@
...
+    from qtpy.QtWidgets import QApplication
+
+    from ._label_components import ICAComponentLabeler
+
+    # get application
+    app = QApplication.instance()
+    if app is None:
+        app = QApplication(["ICA Component Labeler"])
+    gui = ICAComponentLabeler(inst=inst, ica=ica, verbose=verbose)
+    gui.show()
```
mscheltienne, line 30: I think it's missing a `app.exec()` after the `gui.show()` which runs the event loop.

mne_icalabel/gui/_label_components.py
```diff
@@ -0,0 +1,216 @@
...
+    fig = Figure(figsize=(width, height), dpi=dpi)
+    canvas = FigureCanvas(fig)
+    ax = fig.subplots()
+    fig.subplots_adjust(bottom=0, left=0, right=1, top=1, wspace=0, hspace=0)
+    ax.set_facecolor("k")
+    # clean up excess plot text, invert
+    ax.invert_yaxis()
+    ax.set_xticks([])
+    ax.set_yticks([])
+    return canvas, fig
```
mscheltienne, line 81: I think I would define each figure as a widget inheriting from `QtWidgets.QWidget`. Very nice find for `FigureCanvas`, I was not familiar with it and it is a class inheriting from `QWidget`. So how about:

```
class TopomapFig(FigureCanvas):
    """Topographic map widget."""
    
    def __init__(self, width=4, height=4, dpi=300):
        fig = Figure(figsize=(width, height), dpi=dpi)
        ax = fig.subplots()
        fig.subplots_adjust(bottom=0, left=0, right=1, top=1, wspace=0, hspace=0)
        ax.set_facecolor("k")
        # clean up excess plot text, invert
        ax.invert_yaxis()
        ax.set_xticks([])
        ax.set_yticks([])
        super().__init__(fig)
```

The asset is that you can now define methods for each plot, like right-click interaction, zoom, scrolling, ... in the widget class.

mne_icalabel/gui/_label_components.py
```diff
@@ -0,0 +1,216 @@
...
+
+def _make_topo_plot(width=4, height=4, dpi=300):
+    """Make subplot for the topomap."""
+    fig = Figure(figsize=(width, height), dpi=dpi)
+    canvas = FigureCanvas(fig)
+    ax = fig.subplots()
+    fig.subplots_adjust(bottom=0, left=0, right=1, top=1, wspace=0, hspace=0)
+    ax.set_facecolor("k")
+    # clean up excess plot text, invert
+    ax.invert_yaxis()
```
mscheltienne, line 50: Why do we need to invert the Y-axis? Topographic maps matrix are up-side down?

mne_icalabel/gui/_label_components.py
```diff
@@ -0,0 +1,216 @@
...
+
+def _make_ts_plot(width=4, height=4, dpi=300):
+    """Make subplot for the component time-series."""
+    fig = Figure(figsize=(width, height), dpi=dpi)
+    canvas = FigureCanvas(fig)
+    ax = fig.subplots()
+    fig.subplots_adjust(bottom=0, left=0, right=1, top=1, wspace=0, hspace=0)
+    ax.set_facecolor("k")
+    # clean up excess plot text, invert
+    ax.invert_yaxis()
```
mscheltienne, line 64: Y-axis inversion is only for topographic map, no?

mne_icalabel/gui/_label_components.py
```diff
@@ -0,0 +1,216 @@
...
+
+def _make_spectrum_plot(width=4, height=4, dpi=300):
+    """Make subplot for the spectrum."""
+    fig = Figure(figsize=(width, height), dpi=dpi)
+    canvas = FigureCanvas(fig)
+    ax = fig.subplots()
+    fig.subplots_adjust(bottom=0, left=0, right=1, top=1, wspace=0, hspace=0)
+    ax.set_facecolor("k")
+    # clean up excess plot text, invert
+    ax.invert_yaxis()
```
mscheltienne, line 78: Y-axis inversion is only for topographic map, no?

mne_icalabel/gui/_label_components.py
```diff
@@ -0,0 +1,216 @@
...
+    ax.set_xticks([])
+    ax.set_yticks([])
+    return canvas, fig
```
mscheltienne, line 67: For the time-series plot, we need to be able to scroll in time, and probably also zoom on both the X and Y-axis. 
How about maybe looking into the `mne-qt-browser`, maybe we can just load the widgets it has?

mne_icalabel/gui/_label_components.py
```diff
@@ -0,0 +1,216 @@
...
+        plt_grid = QGridLayout()
+        plts = [_make_topo_plot(), _make_ts_plot(), _make_spectrum_plot()]
+        self._figs = [plts[0][1], plts[1][1], plts[2][1]]
+        plt_grid.addWidget(plts[0][0], 0, 0)
+        plt_grid.addWidget(plts[1][0], 0, 1)
+        plt_grid.addWidget(plts[2][0], 1, 0)
+
+        # TODO: is this the correct function to use to render? or nah... since we don't have 3D?
+        self._renderer = _get_renderer(name="ICA Component Labeler", size=(400, 400), bgcolor="w")
+        plt_grid.addWidget(self._renderer.plotter)
```
mscheltienne, line 108: Probably not, I don't think we need this at all.

mscheltienne (commented):
That was not what I had in mind ahah I was thinking of a GUI that loads components one by one, not necessarily from the same `(raw, ica)`. But that fits way better in the API of this package, so I like it!
I'll add a second file with a second iteration to this PR, I got some time today :)

adam2392:
@mscheltienne sorry I haven't gotten around to this yet, but perhaps we can do a pair session this week on getting this up and running?

I pushed up some fixes to at least let it compile.

mscheltienne:
Same it was on my mind for a while.. but I was lacking the time last week. I can work on this on Wednesday or Thursday. 
We can do a pair session if you want.

mne_icalabel/gui/_label_components.py
```diff
@@ -0,0 +1,236 @@
...
+        self.buttonGroup_labels.buttonGroup.setExclusive(True)
+
+    def closeEvent(self, event):
+        """Clean up upon closing the window.
+
+        Check if any IC is not labelled and ask the user to confirm if this is
+        the case. Save all labels in BIDS format.
+        """
+        self.update_saved_labels()
+        print (self.saved_labels)
```
mscheltienne, line 235: @adam2392 Here it is, instead of a print we need to export the saved labels. This event corresponds to "closing the GUI".

adam2392: Hi  @mscheltienne. Adding this now. 

So one issue is that we need the filename to export to BIDS. There are two ways I could see a user doing this:

1. If they load in the raw/ica and it's already in BIDS, instead of passing in the instance, we pass in the BIDS filenames. Then we can extract the BIDS path and save upon closing without requiring the user to pass in a filename because it would just be auto-saved to the "location" it needs to be saved. 
2. If they load in the raw/ica instance, then we need the corresponding filename.

1 is kind of constraining because it requires the users to store the data in exactly the way we need/expect. 2 requires adding an additional UX pipeline for the user to add either the BIDS entities to infer the final filename, or actually add the output filename.

WDYT?

adam2392: Actually, a 3rd option is:

3. If they load in the raw/ica instance, we just modify the ica instance to have labels in the backend. I think, we can populate the `labels_` dictionary of the ICA instance in place: https://mne.tools/stable/generated/mne.preprocessing.ICA.html. Once users close the labeling GUI, then they can run `write_components_tsv` themselves. Or we can add a save-as button.

^ basically, saving automatically to disc when closing the file isn't ideal.

This might require https://github.com/mne-tools/mne-python/issues/9846.

mscheltienne: I'll think about it, for now, I was more inclined towards (1) because it was constrained. Supporting only datasets in BIDS format makes this first version simpler. 

I don't think it's a good idea to have a pop-up GUI to ask the user for a BIDS path, it can be more error-prone and I don't see a reason to save the IC labels in a .tsv sidecar in a BIDS folder if the original data and ICA decomposition are not in the same BIDS folder.

I like option (3) as well, modifying the ICA instance in place makes a lot of sense, and does not entirely make the .tsv redundant as it will also contain the information about who authored the labels (and other information).

adam2392: Yeah the more I think about it...

I'm inclined towards 3) with the pipeline modifying ICA in place, and then expecting the user to save it themselves to BIDS using the API provided in mne-icalabel. The MNE-Python API should eventually support saving the ICA instance with labels, so hypothetically someone might not work in BIDS. 

Optionally, we can also provide a menu inside the GUI to "save-as" and the user can specify the file. 

I think to move this forward, I'll just hack the ICA instance and then provide an example showing how we can save it using the above pipeline. Eventually, we would like to make this pipeline possible via the PR into MNE-Python and FIFF-constants.

mscheltienne: Agree, that makes the most sense.

mne_icalabel/gui/_label_components.py
```diff
@@ -0,0 +1,288 @@
...
+        """Change the component time-series source to plot."""
+        # TODO: change to use ica
+        # TODO: make this work and actually embed in the plot
+        fig = inst.plot(block=False, order=[idx])
+
+        self.axes.set_title("")
+        self.fig = fig
+        self.fig.tight_layout()
+        self.fig.canvas.draw()
+        self.fig.canvas.flush_events()
```
adam2392, line 91: @marsipu this somehow plots the time-series outside the widget. Do you happen to know how to make this work based on your suggestion here: https://github.com/mne-tools/mne-qt-browser/issues/117#issuecomment-1106836014?

Context: I'm trying to add the time-series plot of one ICA source at a time into one of the subplots of the GUI. I would ideally like to just build on the work you have. Rn I'm just plotting the raw inst for testing.

adam2392:
@mscheltienne I'm not sure how to add the time-series plot correctly... so need to figure that out, but the BIDS pipeline I think is outlined now in the example + added code to the `write_components_tsv` function.

mne_icalabel/gui/_label_components.py
mscheltienne: I am surprised by the `axes` and `fig` used below as those are specific to the matplotlib backend(s), and I don't think we have them in the QtBrowser.

mscheltienne:
@adam2392 Ok great, I'll have a look and I'll look into the time-series plot.

mne_icalabel/gui/_label_components.py
```diff
@@ -0,0 +1,278 @@
...
+        del dummy_fig
+
+        # update figures
+        self.widget_topo.redraw()
+        self.widget_psd.redraw()
+        # retrieve layout and swaap timeSeries widget
+        self.central_widget.layout().replaceWidget(
+            self.widget_timeSeries,
+            self._ica.plot_sources(self._inst, picks=[self._current_ic]),
+        )
```
adam2392, line 228: Things to improve...

mscheltienne: Yes, the structure has to be reworked, for now, it's mostly to test it out.

adam2392: Just adding a comment to link to the code in the discourse :p 

https://mne.discourse.group/t/embedding-time-series-plot-of-mne-qt-browser-into-our-own-gui/5090

marsipu:
@adam2392 @mscheltienne I added a comment in the [forum](https://mne.discourse.group/t/embedding-time-series-plot-of-mne-qt-browser-into-our-own-gui/5090/9). I am afraid I currently don't see an easy workaround for the solution you seek. But I also think a possible solution is doable and would be beneficial to the qt- and maybe even for the matplotlib-backend.
For example with the ability to replace data-instances for the plot on a running instance would it also make possible to select e.g. raw-files on the fly from a file-menu in the plot.

adam2392:
> @adam2392 @mscheltienne I added a comment in the [forum](https://mne.discourse.group/t/embedding-time-series-plot-of-mne-qt-browser-into-our-own-gui/5090/9). I am afraid I currently don't see an easy workaround for the solution you seek. But I also think a possible solution is doable and would be beneficial to the qt- and maybe even for the matplotlib-backend.

https://github.com/mne-tools/mne-qt-browser/issues/133#issuecomment-1189574688

@mscheltienne perhaps the best path forward for now is forcing the matplotlib back-end, and using the figure from `ica.plot_sources()`? Then I think it should work as it does for the topograph/psd plot since those use matplotlib?

So we would:
1. set browser backend to matplotlib: https://mne.tools/dev/generated/mne.viz.set_browser_backend.html
2. then take `fig = ica.plot_sources(...)` and attach fig to the TimeSeriesWidget

Does this seem correct to you? Then we can add some basic tests and close out this PR until v2.0 of the GUI.

mscheltienne:
@adam2392 Honestly, I think embedding the `matplotlib` browser is more work. The `matplotlib` one is composed of too many different axis with different interactions. 

> Added the time-series QT Browser, but for now, it works only with the first IC selected, the next ones are plotted in a new window.

I added that very quickly as background during a meeting.. The problem is very simple to fix: `self.widget_timeSeries` has to be overwritten with the new browser widget. For now, it is initialized with the dummy figure widget, swapped to the browser on the first IC selection. and then I forgot to overwrite it, so when a new IC is selected, it tries to swap the dummy figure widget (now non-existent) with a new browser instance. Thus the first browser instance is not replaced and a new browser is opened in a new window.

I'll fix that tomorrow, and it should work OK.

-----------------

@marsipu To embed only the trace/main plotting area without all the interactiveness, scrollbar, etc.. How about just opening a new browser, and then removing all the unwanted widgets from the layout? Sure that's not the most efficient.. but it should work fine and we don't even need to bother with all the signals/slots since the widgets will be just gone.

marsipu:
@mscheltienne I haven't tried that yet, maybe that works. Switching through multiple ICA components and recreating the plot each time might take some time though.

mscheltienne:
Probably, but that will be good enough for a first version until we have a way to update the trace/label.

mscheltienne:
Refactored to improve the structure/code. I'll check if I can figure out how to update the traces from the `pyqtgraph` backend.

adam2392:
@mscheltienne I tried it out, but I keep getting the window to look like this when I run the example script. Not sure if maybe it's my computer only tho. Are you getting the same thing?

Screen Shot 2022-07-21 at 10 40 30 AM

mscheltienne:
Wooow! That's very artistic. No, I can't say I got this behavior. Here is what it looks like on my Windows machine:

https://user-images.githubusercontent.com/73893616/180309359-2ee825ad-2a12-49ea-99a2-057b0c79ab8f.mp4

I tested it on my macOS today, and I had the same behavior. The experience was even a bit better because I had the GUI on a second monitor, so each time I would swap IC the MNE-logo pop-up would open on my main monitor instead of the monitor with the GUI.

--------------

I just noticed that the reset button is gone. I must have messed up something during the refactor and clean-up..

mscheltienne:
I ran it via the entry-point: `mne_gui_ic_annotation --dev` maybe this one works and not the function?
I prefer to use entry-points for Qt GUIs because I use Spyder as an IDE and as it's based on Qt, it already runs an event-loop, so usually, I can't test GUIs through spyder directly.

adam2392:
Wow that looks nice! Thanks for fixing the issue with PyQT.

Dam maybe there's some version issue then or my Mac just sucks... :/

adam2392:
Aaron has been converging with me on a QA workflow for ANTS:

1. filter raw data and visualize it. Remove definite bad channels (that don't look like they could even be recovered by ICA)
2. run ICA on the filtered minus definite bad channels
3. remove bad ICA components
4. re-visualize the reconstructed data
5. repeat for 3-4 subjects

@ayoun25 will also try to get a short 4-5 slide deck of this process for those 3-4 subjects, so we can share w/ you and Anand. Then perhaps if we all agree on the marking standard, he can process all subjects next week w/ the GUI :0.

adam2392:
> 1. filter raw data and visualize it. Remove definite bad channels (that don't look like they could even be recovered by ICA)
> 2. run ICA on the filtered minus definite bad channels
> 3. remove bad ICA components
> 4. re-visualize the reconstructed data
> 5. repeat for 3-4 subjects

Going through this process with Aaron, I realized that sometimes marking ICA isn't necessarily an independent step. Sometimes you're not sure which channels are "definite bad" in the raw data, and maybe this messes up your ICA, so you have to repeat the process.

I think in the future version of the GUI (or MNE-QT-Browser), what we might want is to enable:

1. load the raw data
2. filter and stuff
3. run ICA via GUI
4. visualize output and remove bad components (possibly automatically via our models)
5. visualize output and repeat process if it doesn't look "correct"?

mscheltienne:
Agree with the idea, but I wonder if this GUI-process overlaps with what is already available in MNELAB, I'll have to test it.

mne_icalabel/annotation/bids.py
```diff
@@ -41,17 +42,31 @@ def write_components_tsv(ica: ICA, fname):
...
 
+    # initialize status, description and IC type
+    status = ["good"] * ica.n_components_
+    status_description = ["n/a"] * ica.n_components_
+    ic_type = ["n/a"] * ica.n_components_
+
+    # extract the component labels if they are present in the ICA instance
+    if ica.labels_:
+        for label, comps in ica.labels_.items():
+            this_status = "good" if label != "brain" else "bad"
```
mscheltienne, line 53:
```suggestion
            this_status = "good" if label == "brain" else "bad"
```

Typo?

mne_icalabel/commands/__init__.py
```diff
@@ -0,0 +1,3 @@
+"""Entry-points for mne-icalabel commands."""
+
+from .mne_gui_ic_annotation import main
```
mscheltienne, line 3:
```suggestion
```

Not needed as the full path is given in setup.cfg

mne_icalabel/gui/__init__.py
```diff
@@ -0,0 +1,32 @@
+from mne.preprocessing import ICA
+
+
+def label_ica_components(inst, ica: ICA, show: bool = True, block: bool = False):
+    """Locate intracranial electrode contacts.
```
mscheltienne, line 5:
```suggestion
    """Launch the IC labelling GUI.
```

mne_icalabel/gui/__init__.py
```diff
@@ -0,0 +1,32 @@
...
+
+    Parameters
+    ----------
+    inst : : Raw | Epochs
+        The epochs or raw object.
```
mscheltienne, line 10:
```suggestion
        `~mne.io.Raw` or `~mne.Epochs` instance used to fit the `~mne.preprocessing.ICA` decomposition.
```

requirements_testing.txt
```diff
@@ -25,4 +25,5 @@ scikit-learn
 pandas
 qtpy
 pyvista>=0.32
-pyvistaqt>=0.4
\ No newline at end of file
+pyvistaqt>=0.4
```
mscheltienne: Why do we need `pyvista` and its Qt binding?

adam2392:
Not sure why the docs still aren't building... is qtpy different from pyqt?..

requirements_testing.txt
adam2392: I was just copying code from MME. is pyvista only needed for 3D plots?

mscheltienne: Yes exactly, it's only for 3D visualization and we don't yet have 3D features to display ;)

mscheltienne:
I'll have a look, but the short answer is yes, `qtpy` goal is to provide bindings to `Qt` through any of the following 4 possible Python bindings: `PyQt5`, `PyQt6`, `PySide2`, `PySide6`. It will figure out which one is installed and is available in the environment. It's just an abstraction layer, so you both need `qtpy` and one of the python binding for Qt.
I'm not sure how the `mne-qt-browser` handles it, I don't see one of the 4 python bindings in the requirements. 

It was developed as part of Spyder: https://github.com/spyder-ide/qtpy

requirements_testing.txt
adam2392: Will get rid of it rn

adam2392:
https://output.circle-artifacts.com/output/job/4c1baf98-a737-4203-93de-51707c087700/artifacts/0/dev/auto_examples/label_components.html#sphx-glr-auto-examples-label-components-py

larsoner:
You should just use MNE-Python's `tools/setup_xvfb.sh`, we use it to set up all Qt requirements and an Xvfb for linux CIs

adam2392:
Oh nice! That works, except for some OS/Python combinations it seems. I'm guessing that setting up the GUI to be robust across CI testing is non-trivial...

Does MNE-Python only test GUI on certain OS/Python combos?

larsoner:
We run GUI tests on all OSes and Python versions. On Linux you need xvfb but on Windows and macOS you only need something extra if you need OpenGL support. Regular windowing and matplotlib should be fine on those OSes without any issues I think

adam2392:
> Screen Shot 2022-07-21 at 10 40 30 AM

I still end up getting this weird visual output. If I run things with the entrypoint, then I get this resulting weird visual.

Interestingly, it shows up this way on @ayoun25 Windows machine too, so I wonder if this is a version issue?...

Screen Shot 2022-07-25 at 9 31 43 PM

adam2392:
> We run GUI tests on all OSes and Python versions. On Linux you need xvfb but on Windows and macOS you only need something extra if you need OpenGL support. Regular windowing and matplotlib should be fine on those OSes without any issues I think

Swell! I think now the CI is working! black magic...

adam2392:
Notes from today:

- seems like the plot_sources is plotting twice (so first off set `show=False` where needed), which is fixed if we explicitly instantiate the widget.
- however, then the button clicking is not registering

mscheltienne:
I did not have as much time to dig in as I was hopping, but the issue is that every time you click on a new IC to display, a new browser is opened and the old one is not deleted. It actually stays.. behind.
I tried different ideas with little success:
- layout.removeItem()
- layout.removeWidget()
- browser.close()
- browser.closeEvent()

This issue will be gone once we can open an initial browser and swap the instance displayed via a method supported by the browser, but for now, I still did not find how to delete/close the old browsers. @marsipu by any chance, do you know how from the top of your head?

I'll keep looking.

mscheltienne:
Also, this is not macOS specific, even on my windows PC it does appear (the one I used for the quick screen recording), once you start resizing the window. So I think you are right @adam2392, it showed up on your computer and not on mine because of the resolution/monitor.

mscheltienne:
@adam2392 Give it a shot, I think it's working correctly on my mac. 
My hypothesis is that the C++ reference was not removed, thus the Widget was not removed. Setting the parent to `None` when replacing the browser seems to do the trick:

```
timeSeries_widget = self.ica.plot_sources(self.inst, picks=[self.selected_component])
self._central_widget.layout().replaceWidget(self._timeSeries_widget, timeSeries_widget)
self._timeSeries_widget.setParent(None)
self._timeSeries_widget = timeSeries_widget
```

adam2392:
@mscheltienne nice! Now I get this. I noticed there is no way of getting rid of the plot for the EOG channels. Do we want to keep it in there? If not, we prolly need to PR the `plot_sources` function.

If so, I think we can merge this!

Screen Shot 2022-07-28 at 9 24 34 AM

mne_icalabel/gui/_label_components.py
```diff
@@ -0,0 +1,274 @@
...
+    ica : ICA
+    """
+
+    def __init__(self, inst: Union[BaseRaw, BaseEpochs], ica: ICA, show: bool = True) -> None:
+        ICAComponentLabeler._check_inst_ica(inst, ica)
+        super().__init__()  # initialize the QMainwindow
+        set_browser_backend("qt")  # force MNE to use the QT Browser
+
+        # keep an internal pointer to the instance and to the ICA
+        self._inst = inst
```
mscheltienne, line 40: @adam2392 We could drop the EOG / ECG channels here. We don't need them anyway for this GUI.

adam2392: But this would modify the outside instance in place. Perhaps we leave it for now? It is kind of nice if you have it to compare against time-series traces that look "like it".

mscheltienne: As you want, we can always do a copy before dropping the channels.
It's True that it is a nice piece of information to have to compare against the IC's time-series. The main issue IMO is the size of that browser and especially of the scrollbar/statusbar.

adam2392: Let's just merge this for now for the sake of giving Aaron something to work with. We can modify as we go.

mscheltienne: Sounds good!

adam2392:
@alexrockhill thanks! That's a big PR. 

What's a good summary? :p 

Is it that we can create the GUIs using the abstract interface? It buys us automatically notebook support?

alexrockhill:
> @alexrockhill thanks! That's a big PR.
> 
> What's a good summary? :p
> 
> Is it that we can create the GUIs using the abstract interface? It buys us automatically notebook support?

The summary is that it is free notebook support and implements all the widgets in a way that looks similar between `Qt` and `notebook` with the standard object-oriented design (e.g. `HBox`, `Buttons` etc.).

PR #71, opened 2022-07-18
[ENH] Add topographic map feature via MNE

anandsaini024:
Script to generate the topomap feature.


mne_icalabel/features/topomap.py
```diff
@@ -0,0 +1,144 @@
...
+def get_topo_array(ica, picks="eeg"):
+
+    """Generates a array of scalp topographic plot of size 64*64
+
+    Parameters:
+    ----------
+
+    ica: ICA
+        Instance of MNE ICA Decomposition mne.preprocesssing.ICA
+
```
mscheltienne, line 23:
```suggestion
```

mne_icalabel/features/topomap.py
```diff
@@ -0,0 +1,144 @@
...
+    In lists, channel type strings (e.g., ['meg', 'eeg']) will pick channels
+    of those types, channel name strings (e.g., ['MEG0111', 'MEG2623'] will
+    pick the given channels. Can also be the string values “all” to pick all
+    channels, or “data” to pick data channels. None (default) will pick all
+    channels. Note that channels in info['bads'] will be included if their
+    names or indices are explicitly provided.
+
+    Returns:
+    --------
+
```
mscheltienne, line 34:
```suggestion
```

mscheltienne:
Hey @anandsaini024, for starters, try to remove a bit all the extra-blank lines everywhere; 

and for the test(s), start with a very simple one that just loads a dataset (from the mne testing dataset), runs an ICA decomposition, and runs this function. And you can just start with an assert on the type and shape returned. Then we will think about more advanced tests to ensure it works correctly.

mne_icalabel/features/topomap.py
```diff
@@ -0,0 +1,126 @@
...
+    pos: Info,
+    vmin=None,
+    vmax=None,
+    res=64,
+    outlines="head",
+    image_interp=_INTERPOLATION_DEFAULT,  #'cubic
+    border=_BORDER_DEFAULT,  #'mean'
+    extrapolate=_EXTRAPOLATE_DEFAULT,  #'head'
+):
+    """Generates a topographic map as image
```
mscheltienne, line 68:
```suggestion
    """Generate a topographic map as image.
```

mne_icalabel/features/topomap.py
```diff
@@ -0,0 +1,126 @@
...
+
+    Parameters:
+    -----------
+    data: array of shape (n_channels,)
+        The data values to plot.
+    pos: instance of mne.info
+    vmin: float | callable | None
+        The value specifying the lower bound of the color range.
+    vmax: float | callable | None
```
mscheltienne, line 77:
```suggestion
    vmax : float | callable | None
```

mne_icalabel/features/topomap.py
```diff
@@ -0,0 +1,126 @@
...
+        The value specifying the upper bound of the color range.
+    res: int
+        The resolution of the topomap image (n pixels in eaxh side, for example 64*64 in this case)
```
mscheltienne, line 80:
```suggestion
    res : int
        The resolution of the square topomap image (in pixels).
```

mne_icalabel/features/topomap.py
```diff
@@ -0,0 +1,126 @@
...
+    %(outlines_topomap)s
+    image_interp: str
```
mscheltienne, line 82:
```suggestion
    image_interp : str
```

mne_icalabel/features/topomap.py
```diff
@@ -0,0 +1,126 @@
...
+        The image interpolation to be used. All matplotlib options are
+        accepted.
+    %(border_topomap)s
+    %(extrapolate_topomap)s
+
+    Returns:
+    --------
```
mscheltienne, line 89:
```suggestion
    Returns
    -------
```

mne_icalabel/features/topomap.py
```diff
@@ -0,0 +1,126 @@
...
+    Zi: Array of size (64*64)
+        Topographic map array
```
mscheltienne, line 91:
```suggestion
    Zi : array of size (n_pixels, n_pixels)
        Topographic map array
```

mscheltienne (changes requested):
Let's start with those items, and please add the simple test to just run the function on a testing dataset file. At least this way all the codecov warning will be gone and it will be easier to review the functions.

mne_icalabel/features/topomap.py
```diff
@@ -0,0 +1,126 @@
...
+
+    return topo_array  # topographic map array for all the components (n_components, 64, 64)
+
+
+def topographic_map(
+    data,
+    pos: Info,
+    vmin=None,
+    vmax=None,
+    res=64,
```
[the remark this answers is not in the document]
anandsaini024: I will add this in further modification after simple tests.

mne_icalabel/features/tests/test_features.py
```diff
@@ -0,0 +1,21 @@
...
+    ica = ICA(n_components=5, method="picard")
+    ica.fit(raw)
+    return ica, raw
+
+def test_topomap_feature():
+    """Test scalp topography array generation"""
+    ica, inst = _create_test_inst_and_ica()
+    topo_array = topomap.get_topo_array(ica, picks="eeg")
+    assert isinstance(topo_array, NDArray)
+    assert topo_array.shape == (ica.n_components_, 64, 64)
```
mscheltienne, line 21: Some more conventions, 2 lines between functions ;) If you run from the root of the repository `black mne_icalabel` it will autoformat this kind of style convention.
```suggestion
from mne.datasets import testing
from mne.io import read_raw
from mne.preprocessing import ICA
from numpy.typing import NDArray

from mne_icalabel.features import topomap


def _create_test_inst_and_ica():
    directory = testing.data_path() / "MEG" / "sample"
    raw = read_raw(directory / "sample_audvis_trunc_raw.fif", preload=False)
    raw.pick_types(eeg=True)
    raw.load_data()
    ica = ICA(n_components=5, method="picard")
    ica.fit(raw)
    return ica, raw


def test_topomap_feature():
    """Test scalp topography array generation"""
    ica, inst = _create_test_inst_and_ica()
    topo_array = topomap.get_topo_array(ica, picks="eeg")
    assert isinstance(topo_array, NDArray)
    assert topo_array.shape == (ica.n_components_, 64, 64)
```

mne_icalabel/features/tests/test_features.py
```diff
@@ -0,0 +1,21 @@
...
+    raw.load_data()
+    ica = ICA(n_components=5, method="picard")
+    ica.fit(raw)
+    return ica, raw
+
+def test_topomap_feature():
+    """Test scalp topography array generation"""
+    ica, inst = _create_test_inst_and_ica()
+    topo_array = topomap.get_topo_array(ica, picks="eeg")
+    assert isinstance(topo_array, NDArray)
```
mscheltienne, line 20: That would actually work, but type hints are designed to be used in the declaration of a method/funcrtion/class for the arguments and the returns. 
If you want to check that a variable is of a given type, you should compare directly against it. In this case: `isinstance(topo_array, np.ndarray)`.

mne_icalabel/features/tests/test_features.py
```diff
@@ -0,0 +1,21 @@
...
+from mne_icalabel.features import topomap
+
+def _create_test_inst_and_ica():
+    directory = testing.data_path() / "MEG" / "sample"
+    raw = read_raw(directory / "sample_audvis_trunc_raw.fif", preload=False)
+    raw.pick_types(eeg=True)
+    raw.load_data()
+    ica = ICA(n_components=5, method="picard")
+    ica.fit(raw)
+    return ica, raw
```
mscheltienne, line 14: Nit-picking again: this approach you choose is very good if you have a code snippet that is repeated in multiple tests.
For now, there is only one test in this file, so there is no need to define this 'loading' step in a separate function.

Also, let's suppose we have multiple tests, e.g. 5, instead of calling 5 times `_create_test_inst_and_ica`, it would be better to load only once and fit only once an ICA, because this step is time-consuming and will impact the speed at which tests runs. Thus, usually across the MNE tests, the dataset and any needed fit is done only once, at the same indentation level as the imports at the top of the test file.

Example: https://github.com/mne-tools/mne-python/blob/6955f977ac12612f38ab30277dcfff94315a0463/mne/preprocessing/tests/test_ica.py#L36-L56

All those variables are defined in the module (file) scope and are accessible directly within all the functions defined in that same module (file).

anandsaini024:
This happened when I tried to rename the branch.

mscheltienne:
@anandsaini024 No need to rename the branch, I was asking to rename the test file from `test_features.py` to `test_topomap.py`.

mne_icalabel/features/tests/test_features.py
anandsaini024: Okay, I thought we are going to write one script to test all the features. Will rename it.

mne_icalabel/features/__init__.py
```diff
@@ -0,0 +1,2 @@
+"""Features for the ICLabel"""
+from features import topomap
```
mscheltienne, line 2:
```suggestion
from .topomap import topographic_map
```

Here you need a relative import. You are in the `features` modules (which corresponds to `.`) and you need to import `topographic_map` from `topomap.py`. So `.topomap` -> `topomap.py` file in the current module (`features`).
That will also fix the tests and run them.

mne_icalabel/features/topomap.py
```diff
@@ -0,0 +1,121 @@
...
+        of those types, channel name strings (e.g., ['MEG0111', 'MEG2623'] will
+        pick the given channels. Can also be the string values “all” to pick all
+        channels, or “data” to pick data channels. None (default) will pick all
+        channels. Note that channels in info['bads'] will be included if their
+        names or indices are explicitly provided.
+
+    Returns
+    -------
+    topo_array : np.ndarray of shape (n_components, n_pixels, n_pixels)
+    """
```
mscheltienne, line 29:
```suggestion
def get_topomaps(ica: ICA, picks=None):
    """Generate an array of scalp topographies (n_pixels, n_pixels) for the picked components.

    Parameters
    ----------
    ica : ICA
        Instance of MNE `~mne.preprocesssing.ICA` decomposition.
    %(picks_ica)s ``None`` (default) will pick all sources in the order fitted.

    Returns
    -------
    topomaps : array of shape (n_components, n_pixels, n_pixels)
    """
```

Let's try to be consistent with the names to make the API simpler:
`topomap.py` so let's consider this feature's name as 'topomap'. Let's go for:

- `get_topomap` for single IC -> (n_pixels, n_pixels)
- `get_topomaps` for multiple ICs at once -> (n_components, n_pixels, n_pixels)

Also, the `picks` here should refer to the `picks_ica` to pick ICs: https://github.com/mne-tools/mne-python/blob/e8ded9719de7425c8d6752d2942a6ecba20fd0fc/mne/utils/docs.py#L2289-L2292
So the default to `eeg` and the docstring is wrong as it is not used to pick data channels but ICs. I think it's also missing the logic below to use the picks. Those lines should be adapted:

```
    data = np.dot(
        ica.mixing_matrix_[:, : ica.n_components_].T,
        ica.pca_components_[: ica.n_components_],
    )
```

One more note, in the docstring I changed the ICA parameter line to 
```
Instance of MNE `~mne.preprocesssing.ICA` decomposition.
``` 

which will render as `Instance of MNE ICA decomposition.` with a x-link to the ICA API documentation. The x-link is handled by the backticks, and the tilde `~` removes the parent modules (mne, preprocessing) and only keeps the class/function/method name.

mne_icalabel/features/topomap.py
```diff
@@ -0,0 +1,121 @@
...
+    )
+    # Create an empty array of size (n_components, 64, 64) to fit topo values
+    topo_array = np.zeros((n_components, 64, 64))
+
+    # f, ax = plt.subplots(1, ica.n_components_) #For Visualization
+    for j in range(n_components):
+        topo_ = np.flipud(topographic_map(data[j, :], ica.info))
+
+        # ax[j].imshow(topo_) # For visualization
+
```
mscheltienne, line 43:
```suggestion
```

mne_icalabel/features/topomap.py
```diff
@@ -0,0 +1,121 @@
...
+        # Set NaN values to 0
+        np.nan_to_num(topo_, copy=False)
+
```
mscheltienne, line 46:
```suggestion
```

mne_icalabel/features/topomap.py
```diff
@@ -0,0 +1,121 @@
...
+    topo_array = np.zeros((n_components, 64, 64))
+
+    # f, ax = plt.subplots(1, ica.n_components_) #For Visualization
+    for j in range(n_components):
+        topo_ = np.flipud(topographic_map(data[j, :], ica.info))
+
+        # ax[j].imshow(topo_) # For visualization
+
+        # Set NaN values to 0
+        np.nan_to_num(topo_, copy=False)
```
mscheltienne, line 45:
```suggestion
        np.nan_to_num(topo_, nan=0.0, copy=False)
```

For code readability, explicitly write what nan is converted to.

mne_icalabel/features/topomap.py
```diff
@@ -0,0 +1,121 @@
...
+
+        # Standardize the values
+        topo_array[j, :, :] = topo_ / np.max(np.abs(topo_))
+
+    return topo_array  # topographic map array for all the components (n_components, 64, 64)
+
+
+def topographic_map(
```
mscheltienne, line 53:
```suggestion
def get_topomap(
```

For API consistency, as explained above.

mne_icalabel/features/topomap.py
```diff
@@ -0,0 +1,121 @@
...
+    data,
```
mscheltienne, line 54:
```suggestion
    data: NDArray[float],
```

mne_icalabel/features/topomap.py
```diff
@@ -0,0 +1,121 @@
...
+    pos: Info,
+    vmin=None,
+    vmax=None,
+    res: int = 64,
+    outlines="head",
+    image_interp=_INTERPOLATION_DEFAULT,  #'cubic
+    border=_BORDER_DEFAULT,  #'mean'
+    extrapolate=_EXTRAPOLATE_DEFAULT,  #'head'
```
mscheltienne, line 62:
```suggestion
    image_interp=_INTERPOLATION_DEFAULT,  # 'cubic'
    border=_BORDER_DEFAULT,  # 'mean'
    extrapolate=_EXTRAPOLATE_DEFAULT,  # 'head'
```

Nit-picking, by convention the comments are written with 2 spaces before the `#` (as you did) and one after.
e.g. `code  # comment`.

mne_icalabel/features/topomap.py
```diff
@@ -0,0 +1,121 @@
...
+):
+    """Generate a topographic map as image.
```
mscheltienne, line 64:
```suggestion
    """Generate a scalp topographic map (n_pixels, n_pixels).
```

I'm mostly removing the words 'image' and 'plot' since they relate to visuals while we are just interested in the underlying data array.

mne_icalabel/features/topomap.py
```diff
@@ -0,0 +1,121 @@
...
+
+    Parameters
+    ----------
+    data : array of shape (n_channels,)
+        The data values to plot.
+    pos : instance of `mne.Info`
```
mscheltienne, line 70:
```suggestion
    pos : `mne.Info`
        Instance of `mne.Info` with the montage associated with the (n_channels,) points.
```

mne_icalabel/features/topomap.py
```diff
@@ -0,0 +1,121 @@
...
+    vmin : float | callable | None
+        The value specifying the lower bound of the color range.
+    vmax : float | callable | None
+        The value specifying the upper bound of the color range.
+    res : int = 64
+        The resolution of the square topomap image (in pixels).
```
mscheltienne, line 76:
```suggestion
    res : int = 64
        The resolution of the square topographic map (in pixels).
```

mne_icalabel/features/topomap.py
```diff
@@ -0,0 +1,121 @@
...
+    image_interp=_INTERPOLATION_DEFAULT,  #'cubic
+    border=_BORDER_DEFAULT,  #'mean'
+    extrapolate=_EXTRAPOLATE_DEFAULT,  #'head'
+):
+    """Generate a topographic map as image.
+
+    Parameters
+    ----------
+    data : array of shape (n_channels,)
+        The data values to plot.
```
mscheltienne, line 69:
```suggestion
    data : array of shape (n_channels,)
        The data points used to generate the topographic map.
```

mne_icalabel/features/topomap.py
```diff
@@ -0,0 +1,121 @@
...
+    vmin : float | callable | None
+        The value specifying the lower bound of the color range.
+    vmax : float | callable | None
+        The value specifying the upper bound of the color range.
+    res : int = 64
+        The resolution of the square topomap image (in pixels).
+    %(outlines_topomap)s
+    image_interp : str
+        The image interpolation to be used. All matplotlib options are
+        accepted.
```
mscheltienne, line 80: I think I would just rename this argument to interpolation, or interp, or interp_method.

mne_icalabel/features/topomap.py
```diff
@@ -0,0 +1,121 @@
...
+    %(border_topomap)s
+    %(extrapolate_topomap)s
+
+    Returns
+    -------
+    Zi : Array of size (n_pixels, n_pixels)
```
mscheltienne, line 86:
```suggestion
    topomap : array of size (n_pixels, n_pixels)
```

Please update the variable name in the code accordingly.

mne_icalabel/features/topomap.py
mscheltienne: I also renamed the returned variable to `topomaps`, please update the variable name in the code accordingly.

mne_icalabel/features/topomap.py
```diff
@@ -0,0 +1,121 @@
...
+        Topographic map array
+    """
+    picks = _pick_data_channels(pos, exclude=())  # pick only data channels
+    pos = pick_info(pos, picks)
+    ch_type = _get_channel_types(pos, unique=True)
+
+    if len(ch_type) > 1:
+        raise ValueError("Multiple channel types in Info structure.")
+    elif len(pos["chs"]) != data.shape[0]:
+        raise ValueError("Number of channels in the Info object the data array do not match.")
```
mscheltienne, line 96:
```suggestion
        raise ValueError("The number of channels in the Info object and in the data array do not match.")
```

mne_icalabel/features/topomap.py
```diff
@@ -0,0 +1,121 @@
...
+    else:
+        ch_type = ch_type.pop()
+
+    picks = list(range(data.shape[0]))
+    sphere = np.array([0.0, 0.0, 0.0, 0.095])
+
+    # inferring x,y coordinates form mne.info instance
```
mscheltienne, line 103:
```suggestion
    # inferring (x, y) coordinates form mne.Info instance
```

mne_icalabel/features/topomap.py
```diff
@@ -0,0 +1,121 @@
...
+    pos = _find_topomap_coords(pos, picks=picks, sphere=sphere)
+    extrapolate = _check_extrapolate(extrapolate, ch_type)
+
+    if len(data) != len(pos):
+        raise ValueError(
+            "Data and pos need to be of same length. Got data of "
+            "length %s, pos of length %s" % (len(data), len(pos))
+        )
```
mscheltienne, line 111: This check looks like a duplicate of the one above (ValueError: The number of channels ...), or am I missing something?

mne_icalabel/features/topomap.py
```diff
@@ -0,0 +1,121 @@
...
+
+    norm = min(data) >= 0
+    vmin, vmax = _setup_vmin_vmax(data, vmin, vmax, norm)
+
+    # interpolation
+    outlines = _make_head_outlines(sphere, pos, outlines, (0.0, 0.0))
+    extent, Xi, Yi, interp = _setup_interp(pos, res, image_interp, extrapolate, outlines, border)
+    interp.set_values(data)
+    Zi = interp.set_locations(Xi, Yi)()
+    return Zi  # Topographic map array of size (64*64)
```
mscheltienne, line 121:
```suggestion
    return interp.set_locations(Xi, Yi)()  # Zi, topomap of shape (n_pixels, n_pixels)
```

Simpler.

mscheltienne (changes requested):
Please have a look at the comments, it's almost done ;)
The biggest point to address is the `picks` argument in the `get_topo_array` (renamed to `get_topomaps`) function which does not relate to data channels but to ICs. Currently, neither the docstring or code are correct in that regard.

mne_icalabel/features/__init__.py
mscheltienne: With the new names, it would be 

```suggestion
from .topomap import get_topomap, get_topomaps  # noqa: F401
```

I'm also adding the `noqa` flag to disable the F401 error raised by static code analysis tool. c.f. https://www.flake8rules.com/rules/F401.html

mne_icalabel/features/tests/__init__.py
```diff
@@ -0,0 +1 @@
+""" Test Features"""
```
mscheltienne, line 1:
```suggestion
```

mne_icalabel/features/tests/test_topomap_feature.py
```diff
@@ -0,0 +1,20 @@
+from mne.datasets import testing
+from mne.io import read_raw
+from mne.preprocessing import ICA
+from numpy.typing import NDArray
+
+from mne_icalabel.features.topomap import get_topomaps
```
mscheltienne, line 6:
```suggestion
from mne_icalabel.features import get_topomap, get_topomaps
```

mne_icalabel/features/tests/test_topomap_feature.py
```diff
@@ -0,0 +1,20 @@
+from mne.datasets import testing
```
mscheltienne, line 1:
```suggestion
import numpy as np
from mne.datasets import testing
```

mne_icalabel/features/tests/test_topomap_feature.py
```diff
@@ -0,0 +1,20 @@
...
+raw.load_data()
+ica = ICA(n_components=5, method="picard")
+ica.fit(raw)
+
+
+def test_topomap_feature():
+    """Test scalp topography array generation"""
+    topo_array = get_topomaps(ica, picks="eeg")
+    assert isinstance(topo_array, NDArray)
+    assert topo_array.shape == (ica.n_components_, 64, 64)
```
mscheltienne, line 20:
```suggestion
def test_topomap_defaults():
    """Test scalp topography array generation"""
    topomaps = get_topomaps(ica, picks="eeg")
    assert isinstance(topomaps, np.ndarray)
    assert topomaps.shape == (ica.n_components_, 64, 64)
    
    # test single topo with fake data
    data = np.random.randint(1, 10, len(raw.ch_names))
    topomap = get_topomap(data, raw.info)
    assert isinstance(topomap, np.ndarray)
    assert topomap.shape == (64, 64)
```

mne_icalabel/features/tests/test_topomap_feature.py
```diff
@@ -0,0 +1,20 @@
+from mne.datasets import testing
+from mne.io import read_raw
+from mne.preprocessing import ICA
+from numpy.typing import NDArray
```
mscheltienne, line 4:
```suggestion
```

mne_icalabel/features/topomap.py
```diff
@@ -0,0 +1,114 @@
...
+    image_interp : str
+        The image interpolation to be used. All matplotlib options are
+        accepted.
+    %(border_topomap)s
+    %(extrapolate_topomap)s
+
+    Returns
+    -------
+    topomap : array of size (n_pixels, n_pixels)
+        Topographic map array
```
mscheltienne, line 86:
```suggestion
        Topographic map array.
```

mscheltienne (changes requested):

mscheltienne:
@anandsaini024 I pushed a couple more style changes. It looks like the error in the test is related to the `picks`.

mscheltienne:
Depends on #75 to fix the `requires_version` decorator and to add the `fill_doc` and `copy_doc` decorators.

mne_icalabel/features/topomap.py
```diff
@@ -0,0 +1,141 @@
...
+    outlines = _make_head_outlines(sphere, pos, outlines, (0.0, 0.0))
+    extent, Xi, Yi, interp = _setup_interp(pos, res, image_interp, extrapolate, outlines, border)
+    interp.set_values(data)
+    return interp.set_locations(Xi, Yi)()  # Zi, topomap of shape (n_pixels, n_pixels)
+
+
+def _check_mne_version():
+    """Check that MNE version is above 1.1."""
+    if not check_version("mne", "1.1"):
+        raise RuntimeError("Topographic feature is only available for MNE ≥ 1.1")
```
adam2392, line 141: Should this be in utils, and maybe add an argument to make it generalize for any version?

mne_icalabel/features/topomap.py
```diff
@@ -0,0 +1,141 @@
...
+from mne.preprocessing import ICA
+from mne.utils import check_version
+from mne.viz.topomap import _check_extrapolate, _make_head_outlines, _setup_interp
+from numpy.typing import NDArray
+
+from ..utils._docs import fill_doc
+
+
+@fill_doc
+def get_topomaps(
```
adam2392, line 28: It seems like `get_topomaps` works on ICA instance, whereas `get_topomap` works on a np.ndarray. Perhaps in line with the MNE tfr module naming, we can make it more explicit like:

`get_topomaps` stays the same
`get_topomap` renamed to something like `get_topomap_array` to make it explicit that it works on an array?

Not wedded to this idea, so just a thought.

adam2392 (commented):
I left a few comments, but overall this LGTM once we address those.

mne_icalabel/features/topomap.py
mscheltienne: Usually, I would say yes, but this code should be removed once we bump the MNE requirement from 1.0 to 1.1. 
I think keeping this 'temporary' code in the same file makes this future step easier, even if in this case I don't foresee this req. bump to happen anytime soon.

I don't think we need a general case for this function, we won't use it anyway and we should support all the versions of MNE included in the requirements.

By the way, this topographic feature requires MNE 1.1 because of recent changes and bug fixes in the interpolation. c.f. https://github.com/mne-tools/mne-python/pull/10579 and https://github.com/mne-tools/mne-python/pull/10617

mne_icalabel/features/topomap.py
mscheltienne: Done, makes more sense indeed!

mne_icalabel/features/topomap.py
adam2392: Okay fair. I'm in favor of just automatically going to only support v1.1+ for mne-icalabel v0.3 since we're in super-fast dev mode.

Maybe add a `#TODO` comment then to remove this when v1.1 is released?

mne_icalabel/features/topomap.py
```diff
@@ -0,0 +1,141 @@
...
+    extrapolate = _check_extrapolate(extrapolate, ch_type)
+
+    # interpolation, valid only for MNE ≥ 1.1
+    outlines = _make_head_outlines(sphere, pos, outlines, (0.0, 0.0))
+    extent, Xi, Yi, interp = _setup_interp(pos, res, image_interp, extrapolate, outlines, border)
+    interp.set_values(data)
+    return interp.set_locations(Xi, Yi)()  # Zi, topomap of shape (n_pixels, n_pixels)
+
+
+def _check_mne_version():
```
adam2392, line 138:
```suggestion
# TODO: remove this and the corresponding import when mne v1.1 is released
def _check_mne_version():
```

adam2392 (approved):

mne_icalabel/features/topomap.py
mscheltienne: That works for me!

mscheltienne (approved):
LGTM after a couple more small changes ;) I'll merge if CIs are green.

mscheltienne:
Thanks @anandsaini024 ! All done for this one, you can now update the branch from the PSD feature PR.

PR #73, opened 2022-07-19
[ENH] Add PSD feature via MNE

anandsaini024:
PSD feature creation script.


adam2392:
Just to clarify, how are these and the topographic PR different from what is included in ICLabel?

I might've missed this discussion.

mscheltienne:
@adam2392 The ICLabel ones are a copy/paste of the MATLAB features.. which are not super clean and nice.
For instance, the ICLabel `rpsd` feature is supposed to have a random component right? Well.. in practice that's not the case because it does something along the lines of:

```
some_2d_matrix = [..., ...]
subsets = [indices of all columns]  # instead of an actual subset
select_and_shuffle = shuffle(some_2d_matrix [:, subsets])  # shuffle the columns
average = avg(select_and_shuffle, axis=1)  # avg across columns
```

And that's not the only example of 'weird' coding choices/behaviors. I added comment in the `features.py` file describing those items when I ran into them, e.g. https://github.com/mne-tools/mne-icalabel/blob/26a2f1fb2829d7f3ee7019c6591775be33ecde98/mne_icalabel/iclabel/features.py#L375-L378

which corresponds to the `pct_data` variable in this file https://github.com/sccn/ICLabel/blob/d0cfb0f8638cd4518ee2e66a3bff0bb32ff40969/eeg_autocorr_welch.m#L1-L25

variable which is never provided to the function https://github.com/sccn/ICLabel/blob/d0cfb0f8638cd4518ee2e66a3bff0bb32ff40969/ICL_feature_extractor.m#L77

Or this one: https://github.com/mne-tools/mne-icalabel/blob/26a2f1fb2829d7f3ee7019c6591775be33ecde98/mne_icalabel/iclabel/features.py#L426-L429

---------------------------

So I asked Anand to prepare functions to extract features from Instance/ICA decomposition using MNE, starting with the topographic map and the PSD based on the `ica.plot_properties` method. The idea is to use those functions to extract features for future models.

mne_icalabel/features/psd.py
```diff
@@ -0,0 +1,111 @@
...
+        reject_by_annotation=reject_by_annotation,
+        verbose=None,
+    )
+    if dB:
+        np.log10(np.maximum(psds, np.finfo(float).tiny), out=psds)
+        psds *= 10
+        psds_mean = psds.mean(axis=0)
+    else:
+        psds_mean = psds.mean(axis=0)
+    return psds_mean  # psds_mean = (n_components, n_freqs)
```
mscheltienne, line 111:
```suggestion
    return psds_mean  # psds_mean = (n_components, n_freqs)
    
```

More and more tiny nit-pick: per convention, usually each file ends with a blank line. Most editor have the following options in the settings:
- Remove spaces from end of line
- Trim extra-empty lines at the end of files
- Add an empty line at the end of files
Those settings apply upon saving.

mne_icalabel/features/psd.py
```diff
@@ -0,0 +1,111 @@
+import numpy as np
+from mne.io.pick import _picks_to_idx
+from mne.time_frequency.psd import psd_multitaper
+from mne.utils import _check_option
+from mne.viz.ica import _prepare_data_ica_properties
+
+
+def get_psd(
```
mscheltienne, line 8:
```suggestion
def get_psds(
```

I would rename this one with an extra `s` to match the naming schemes of the topomap feature.

mne_icalabel/features/psd.py
```diff
@@ -0,0 +1,111 @@
...
+    ica : instance of mne.preprocessing.ICA
+        The ICA solution.
+    inst : instance of Epochs or Raw
+    picks : str | list | slice | None
+        In lists, channel type strings (e.g., ['meg', 'eeg']) will pick channels
+        of those types, channel name strings (e.g., ['MEG0111', 'MEG2623'] will
+        pick the given channels. Can also be the string values “all” to pick all
+        channels, or “data” to pick data channels. None (default) will pick all
+        channels. Note that channels in info['bads'] will be included if their
+        names or indices are explicitly provided.
```
mscheltienne, line 40: Same comment as for the topomap, this `picks` argument does not refer to the data channels but to the ICs.

adam2392:
FYI to fix the style easily and the sorting of imports:

`black ./mne_icalabel/*` and `isort ./mne_icalabel` commands, or type `make black` and `make isort`. The other checks like flake8 (style), pydocstyle, and mypy can be fixed manually.

doc/whats_new.rst
```diff
@@ -35,6 +35,7 @@ API
 ~~~
 
 - Add topographic feature using MNE with `~mne_icalabel.features.get_topomap_array` and `~mne_icalabel.features.get_topomaps` by `Anand Saini`_ and `Mathieu Scheltienne`_ (:gh:`71`)
+- Add psd feature using MNE with `~mne_icalabel.features.get_psds` by `Anand Saini`_ (:gh:`73`)
```
mscheltienne, line 38:
```suggestion

- Add PSD feature using `mne.time_frequency.psd_multitaper` with `~mne_icalabel.features.get_psds` by `Anand Saini`_ (:gh:`73`)
```

mne_icalabel/features/psd.py
```diff
@@ -0,0 +1,118 @@
+from typing import Optional, Union
+
+import numpy as np
+from mne import BaseEpochs
+from mne.io import BaseRaw
+from mne.io.pick import _picks_to_idx
+from mne.preprocessing import ICA
+from mne.time_frequency.psd import psd_multitaper
```
mscheltienne, line 8:
```suggestion
from mne.time_frequency import psd_multitaper
```

mne_icalabel/features/psd.py
```diff
@@ -0,0 +1,118 @@
...
+    dB: bool = True,
+):
+    """Compute the power spectral density of the ICA components.
+
+    Parameters
+    ----------
+    ica : instance of mne.preprocessing.ICA
+        The ICA solution.
+    inst : instance of Epochs or Raw
+        The data to use in calculating PSDs.
```
mscheltienne, line 42:
```suggestion
    inst : instance of Epochs or Raw
        `~mne.io.Raw` or `~mne.Epochs` instance used to fit the `~mne.preprocessing.ICA` decomposition.
```

mne_icalabel/features/psd.py
```diff
@@ -0,0 +1,118 @@
...
+
+    Returns
+    -------
+    psds_mean : ndarray, shape(n_channels, n_freqs)
+         Mean of power spectral densities on channel.
+    """
+    if picks is None:  # plot all components
+        picks = range(0, ica.n_components_)
+    else:
+        picks = _picks_to_idx(ica.n_components_, picks)
```
mscheltienne, line 89:
```suggestion
    picks = _picks_to_idx(ica.n_components_, picks)
```

You can provide `picks=None` to `_picks_to_idx` directly.

mne_icalabel/features/psd.py
```diff
@@ -0,0 +1,118 @@
...
+    kind, dropped_indices, epochs_src, data = _prepare_data_ica_properties(
+        inst, ica, reject_by_annotation=True, reject="auto"
+    )
+    Nyquist = inst.info["sfreq"] / 2.0
+    fmax = min(inst.info["lowpass"] * 1.25, Nyquist)
```
mscheltienne, line 94:
```suggestion
    nyquist = inst.info["sfreq"] / 2.0
    fmax = min(inst.info["lowpass"] * 1.25, nyquist)
```

Nit-picking.

mne_icalabel/features/psd.py
```diff
@@ -0,0 +1,118 @@
...
+        reject_by_annotation=reject_by_annotation,
+        verbose=None,
+    )
+    if dB:
+        np.log10(np.maximum(psds, np.finfo(float).tiny), out=psds)
+        psds *= 10
+        psds_mean = psds.mean(axis=0)
+    else:
+        psds_mean = psds.mean(axis=0)
+    return psds_mean  # psds_mean = (n_components, n_freqs)
```
mscheltienne, line 118:
```suggestion
    if dB:
        np.log10(np.maximum(psds, np.finfo(float).tiny), out=psds)
        psds *= 10
    return psds.mean(axis=0)  # (n_components, n_freqs)
```

mne_icalabel/features/tests/test_psd.py
```diff
@@ -0,0 +1,19 @@
...
+from mne.io import read_raw
+from mne.preprocessing import ICA
+
+from mne_icalabel.features import get_psds
+
+directory = testing.data_path() / "MEG" / "sample"
+raw = read_raw(directory / "sample_audvis_trunc_raw.fif", preload=False)
+raw.pick_types(eeg=True)
+raw.load_data()
+ica = ICA(n_components=7, method="picard")
```
mscheltienne, line 12: Is there a reason you asked for 7 components? Else let's lower it to 5 as it will slightly speed up the fit and the test.

mscheltienne (changes requested):

doc/whats_new.rst
mscheltienne: Entries in the changelog are separated by one empty line, else the documentation build will complain ;)

mne_icalabel/features/tests/test_psd.py
anandsaini024: No, I was testing something. We can set it to 5.

mne_icalabel/features/psd.py
```diff
@@ -0,0 +1,113 @@
...
+        or ``'power'``. If ``'power'``, the power spectral density is returned.
+        If ``output='complex'``, the complex fourier coefficients are returned
+        per taper.
+    dB : bool
+        Whether to plot spectrum in dB. Defaults to True.
+
+    Returns
+    -------
+    psds_mean : ndarray, shape(n_channels, n_freqs)
+         Mean of power spectral densities on channel.
```
mscheltienne, line 85:
```suggestion
    psds : array of shape (n_channels, n_freqs)
         The independent component power spectral density.
```

The mean is across epochs, right? So not across channels.

doc/api.rst
```diff
@@ -57,6 +57,7 @@ Contains functions to extract features from `~mne.preprocessing.ICA` instance an
 
    get_topomap_array
    get_topomaps
+   get_psds
```
adam2392, line 60: Do we want a `get_psds_array`

mne_icalabel/features/tests/test_psd.py
```diff
@@ -0,0 +1,19 @@
...
+raw.pick_types(eeg=True)
+raw.load_data()
+ica = ICA(n_components=5, method="picard")
+ica.fit(raw)
+
+
+def test_psd_feature():
+    """Test scalp topography array generation"""
+    psds_mean = get_psds(ica, raw)
+    assert isinstance(psds_mean, np.ndarray)
```
adam2392, line 19: Can we add a better test perhaps? Anything that can be inspired from MNE-Python test suite?

doc/api.rst
mscheltienne: I am more inclined to make `get_topomap_array` private and remove it from the documentation. `get_topomaps` computes the topographic map one by one by calling `get_topomap_array`.
With the PSD feature, this is not needed and all PSDs can be computed at once. 

I don't see the added value of a function that can take a single IC as a numpy array and compute the corresponding feature. And supporting only the MNE API (with Raw|Epochs|ICA) gives us better control over what is provided as input to those functions than a numpy array.

mne_icalabel/features/tests/test_psd.py
mscheltienne: Yes definitely.

doc/api.rst
adam2392: Okay agreed.

adam2392: @anandsaini024 shall you then make the `get_topomap_array` a private function and remove it from `api.rst`

mscheltienne: Done.

mscheltienne:
@anandsaini024 Please look into additional tests as we discussed:
- test inspired from tests of the PSD functions in MNE
- test on signals with known frequency content, e.g. sinusoid, amplitude modulated signal, ...
- test with different inputs for the arguments of `get_psds`

I'll try to double-check this PR this week.

mne_icalabel/features/tests/test_psd.py
```diff
@@ -0,0 +1,56 @@
...
+    with pytest.raises(ValueError):
+        get_psds(ica, inst, picks="eeg")
+    with pytest.raises(ValueError, match="All picks must be < n_channels"):
+        get_psds(ica, inst, picks=10)
+    with pytest.raises(ValueError):
+        get_psds(ica, inst, fmax=[210, 410])
+    with pytest.raises(ValueError):
+        get_psds(ica, inst, fmax="string")
+    with pytest.raises(ValueError):
+        get_psds(ica, inst, normalization="string")
```
mscheltienne, line 56: Please add the `match` keyword and either part of the error message, the entire error message, or a regex pattern that matches the error. This will avoid catching ANY `ValueError`.

```
with pytest.raises(Error, match="error I want to test against"):
    function_that_raises_that_error()
```

Also, if you notice that an error message is not super clear, then you can start checking the user input at the beginning of the function. I've done it for `fmin` and `fmax`, but any error message that does not give clear guidelines to the user on how to correct it is not good enough.

mscheltienne (commented):
@anandsaini024 Please check that the output you are getting from the function makes sense also when you vary the different arguments. 
For instance, what happens when you provide a raw which does have `"bad"` annotations and `reject_by_annotation` was set to True? Or to False?
And so on.. please test all the arguments in different scenarios to make sure they each work. Those tests that you will do manually at the beginning can then be incorporated in `test_psd_arguments`.

mmagnuski:
Hi, I popped in to see whats new in mne-icalabel (great project!) and saw your discussion of `pct_data` variable and how it is not used in the Matlab code. I think (but I may be wrong) that it is related to how EEGLAB usually computed PSD - to save time the spectrum was computed only for a portion of the data (especially important for very long raw recordings and the fact to you expect instant feedback when working via GUI). This option was controlled with a special parameter that in research usage would be set to 100. I think this may be the reason for the behavior of `pct_data` you see.

mscheltienne:
Maybe, I didn't dig in too far, but it looked to me like the feature extraction code I translated was half-baked. 
And I think the unused `pct_data` variable appears both in the PSD and autocorrelation features. 

Also, I'll finalize this PR at some point ;)

mscheltienne:
This is now too out of date, especially with the added `Spectrum` objects. Closing.

PR #75, opened 2022-07-21
[MNT] Move utils to a separate module and add utilities for docstring and testing

mscheltienne:
Move `utils.py` to `utils._checks.py` since it only contained one function to check the ICA and the instance provided.
Add `fill_doc` with keys taken form the MNE docdict and from our own docdict.
Add `copy_doc`, similar to the MNE decorator.
Add `requires_version` that uses `mne.utils.check_version` instead of `mne.utils._testing.check_version`.


adam2392:
Nice I like it, and then we can replace in a lot of the docstrings:

`<param_name> : description` with `%(param_name)s` right?

mscheltienne:
@adam2392 Exactly, and also use all the docstring already defined in the MNE docdict: https://github.com/mne-tools/mne-python/blob/main/mne/utils/docs.py

Deduplicating docstrings is such a nice feature :) I'm already using this code in many repositories, it works really well. The only part not covered is the `@verbose` decorator which also acts as a docstring filler in MNE. If we want to use a similar decorator, we will have to copy/paste it in our `_docs.py`, else it will use the docdict from MNE only. For now, I don't see the need as the only call to `@verbose` is in https://github.com/mne-tools/mne-icalabel/blob/26a2f1fb2829d7f3ee7019c6591775be33ecde98/mne_icalabel/datasets/icalabel/icalabel.py#L18-L46 

which does not fill any `mne-icalabel` specific docdict entries.

---------

And for the testing part, #71 needs a different `requires_version` decorator than `mne.utils.requires_version`, c.f. https://github.com/mne-tools/mne-python/issues/10941

mscheltienne:
@adam2392 `mypy` is kind of a pain.. Do you know how to fix or ignore the remaining issues?

adam2392:
Yes I'll fix it.

mscheltienne:
Found it by chance, those error messages were a bit too cryptic! @adam2392 LGTM, merge if happy.

adam2392 (approved):

PR #82, opened 2022-07-31
Update minimum python version

TomDonoghue:
Thanks for contributing. If this is your first time,
make sure to read the [contributing guide.](https://github.com/mne-tools/mne-python/blob/main/CONTRIBUTING.md)

PR Description
--------------

The tags and installation description imply that this package requires a minimum version of py 3.7 (and works on 3.8, 3.9, etc), however, the listed `python_requires` tag (which gets listed on the PYPI site) instead lists the requirement as `~= 3.7`, This quick fix updates that listing. 

Merge checklist
---------------

- [ ] All comments resolved
- [ ] This is not your own PR
- [ ] All CIs are happy

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -33,7 +33,7 @@ project_urls =
   Source = https://github.com/mne-tools/mne-icalabel
 
 [options]
-python_requires = ~= 3.7
+python_requires = >= 3.7
 install_requires =
   mne >= 1.0
   numpy >= 1.16.0


mscheltienne:
Thanks for the contribution. I only learned recently about it as well, this syntax is called a "Compatible release clause". See https://peps.python.org/pep-0440/#version-specifiers

`~= 3.7` means `3.7 <= X <= 4`

In this case, it's equivalent to `>=3.7`. See:

> For example, the following groups of version clauses are equivalent:

```
~= 2.2
>= 2.2, == 2.*

~= 1.4.5
>= 1.4.5, == 1.4.*
```


Or is there something else I'm missing?

TomDonoghue:
Ohh, the "~=" operator works slightly differently than I thought - I was reading "~= 3.7" as being tagged to any sub-version of 3.7 (I was reading it as what would actually be "~= 3.7.0" or similar, I think). Basically, the way ~= gets parsed matches one fewer version number than I realized, such that this basically tags the python version to something like '>= 3.7, < 4)', as another way to writing '>=3.7, ==3.*', but doesn't restrict to 3.7. 

In that case, this does work / is consistent with the tags, I just got confused. In practice (given there is no Python 4), this usage does mean the same as  '>=3.7'.

mscheltienne:
Exactly. Overall, the `~=` operator is slightly safer, even if in this case Python 4.0 is nowhere to be seen on the horizon ;)

PR #88, opened 2022-08-01
[MNT] Improve packaging, managing of dependencies, and managing of ignored warning/errors

mscheltienne:
Looking through the different configuration files for the different automatic checker/formatter/tests.. it's a bit messy and most of the stuff copy/pasted from MNE is probably not required here. Especially, I'm a bit worried by all the different ignore-clause which could mask issues.

So my plan is to try to simplify everything and see how it goes, and while I'm at it let's use `pyproject.toml` to its full potential. After the discussion in #49, I started playing around with the different packaging approaches, let's see if I can get something simpler here ;)


mscheltienne:
@adam2392 Soo, I changed quite a few things. I'll sum it up below, let me know your thoughts about those changes:
(apologies, that's a long post..)

-----------------------------

### Packaging - Configuration

Previously pacakge and tool configurations were handled in both `setup.cfg` and `pyproject.toml`. Instead, I moved everything to `pyproject.toml`. The goal is to get all package and tool configurations in the same file. The only 2 exceptions are:
- codespell, the list of ignored words (moved to the file `.codespellignore`, default naming). For now, it only contains one word.
- flake8, which does not yet support `pyproject.toml`. The conversation in https://github.com/PyCQA/flake8/issues/234 has been locked more than a year ago and does not make any progress, with subsequent PR to add supports for `pyproject.toml` being marked as spam.

`setup.cfg` has been removed. `setup.py` is still present. `setuptools` is a bit late in implementing PEP660 (currently in beta, it will be available soon) compared to other packaging tools (e.g. `poetry`) and thus still requires a legacy `setup.py` to install a package in editable mode. Note this limitation is only for editable mode.

### Packaging - Files included in the wheel and the tarball

The inclusion/exclusion patterns are now set to https://github.com/mscheltienne/mne-icalabel/blob/24896bd77697d1cd4d8064bc62b0b6d4bbe6079a/pyproject.toml#L120-L122 which includes all the modules/files in `mne_icalabel` except the `tests` modules.

Package data, such as the model weights, which are inside `mne_icalabel` are defined here https://github.com/mscheltienne/mne-icalabel/blob/24896bd77697d1cd4d8064bc62b0b6d4bbe6079a/pyproject.toml#L124-L125 instead of in a `MANIFEST.in`.

This brings me to `MANIFEST.in`, for now, I removed it. I don't see why we need to include anything outside `mne_icalabel` in the tarball. If you want to include some of the files outside the package (i.e. at the root of the repository) in the tarball, let me know. I know that some people like to package in the tarball the documentation build (`doc`, `examples`), the configuration file (`pyproject.toml`, `.flake8`, `.codespellignore`, ...)

### Dependencies

The base dependencies are listed here https://github.com/mscheltienne/mne-icalabel/blob/24896bd77697d1cd4d8064bc62b0b6d4bbe6079a/pyproject.toml#L39-L47 and include 2 additional dependencies: `pooch` which is imported in the `datasets` module and was missing, `importlib-metadata` which is a compatibility fix for python 3.7.

The optional dependencies are now listed here https://github.com/mscheltienne/mne-icalabel/blob/24896bd77697d1cd4d8064bc62b0b6d4bbe6079a/pyproject.toml#L49-L107 and are splitter in different keys depending on what we need. I also removed all the dependencies that were not used, except for the `doc` key (still a WIP).

Now, you can install the package with all the requirements with `pip install -e .[all]` (local clone) or with `pip install mne-icalabel[all]`. The key `[full]` is equivalent to `[all]`.
For the GUI and for MNE's full ICA decomposition capabilities: `pip install mne-icalabel[ica,gui]`.

### CIs

I reworked the unit-tests CI and removed what was not needed, e.g. installing `PyQt6 PyQt6-sip PyQt6-Qt6` to run the build.
The `style` test uses GitHub actions except for `pydocstyle`.
The `build` test now runs on all 3 OS.
The `test` test now caches the MNE and MNE-ICALabel testing dataset (key base on the dataset version) to avoid re-downloading the dataset every time if it's already available. Instead of the old `setup_xvfb.sh` script, the `qt5-default` are installed on linux here https://github.com/mscheltienne/mne-icalabel/blob/24896bd77697d1cd4d8064bc62b0b6d4bbe6079a/.github/workflows/unit_tests.yml#L124-L128 and the tests are run normally on macOS or windows, or via Xvfb on linux here https://github.com/mscheltienne/mne-icalabel/blob/24896bd77697d1cd4d8064bc62b0b6d4bbe6079a/.github/workflows/unit_tests.yml#L153-L156
The choice of `Qt5` is limited by MNE which imports from `PyQt5` in version 1.0.0 and below, and from `qtpy` since version 1.1. https://github.com/mne-tools/mne-python/blob/cee9e96e99abc98ce4fd0ae5e0a9ffcb0c3919f3/mne/viz/backends/_utils.py#L97-L122 
Also, the coverage is now uploaded only by one test.

### Pytest - `conftest.py`

We actually used nothing from this file.. and it was masking an issue with 3 missing `__init__.py` files.
The 2 fixtures defined were for tests creating `matplotlib` figures, and we don't have any. The only use of `matplotlib` is in the GUI, and uses a Qt interactive backend embedded in the GUI (i.e. not the `Agg` backend). 

### Future testing of the GUI

For the GUI, future tests will probably require `pytest-qt` which let us fake actions on the GUI (e.g. pressing a button). c.f. `mnelab`.

### MNE 1.1

I will also include MNE 1.1 in this PR when it's out, and I'll drop support of MNE 1.0.
That means we can adapt some of the points above, like using Qt6 instead of Qt5.

-----------------------------

There are also some other minor changes in the configuration, like matching the `flake8`, `black`, ... line-length (I think to 100 characters). 

The next steps are:

- [ ] Fix the doc build CI (Oopsie, I broke something)
- [ ] Check the doc extra-dependencies
- [ ] Check the release CI
- [ ] Update the make file
- [ ] Drop support for MNE 1.0 and bump requirement to 1.1

adam2392:
This mostly looks good to me! I'll take a deeper look. I'm okay with dropping the Manifest if not needed.

The docs build is broken cuz I think you removed some requirements files. Perhaps we should also update the installation page with the new install options as well via pip? And then what would conda install?

mscheltienne:
When you have a bit of time, do take a look. I think it's easier to have a look directly at the [PR branch](https://github.com/mscheltienne/mne-icalabel/tree/setup) and at the files remaining than to look at the file change tab ;)

For the doc build, it's complaining before installing the requirements. Apparently:

```
./scripts/circle_ci_dependencies.sh: No such file or directory
```

I'll fix it, slowly and carefully as I'm not as familiar with Circle as I am with Azure or GitHub.

-----------

Yes, when everything is clean-up and tidy-up, I'll also update the installation page.

Very good question with `conda`, I have no idea how it handles the optional dependency and keys. Looking at the MNE recipe https://github.com/conda-forge/mne-feedstock/blob/main/recipe/meta.yaml you can define several `conda install xxx`, e.g. `mne-base` or `mne`.
I'm guessing we can go with `conda install -c conda-forge mne-icalabel` for full functionalities (e.q. to `pip install mne-icalabel[gui,ica]`) and with an extra key, e.g. `conda install -c conda-forge mne-icalabel-dev` for all the dev. related extra dependencies?

adam2392:
Took an actual look! And it looks great! Like that it's so clean now.

For the release, do we need to still add the PYPI secrets? Idr.. But it looks solid!

mscheltienne:
For Pypi, usually, you use a token. This is how I set it up on other projects on GitHub actions:

```
      - name: Build and publish
        env:
          TWINE_USERNAME: __token__
          TWINE_PASSWORD: ${{ secrets.PYPI_API_TOKEN  }}
        run: |
          python -m build
          twine upload dist/*
```

adam2392:
Okay looks like one of us had added it into this repo before. I see it in the settings

adam2392:
For circleCi somehow the installation of mne-icalabel is not being registered...?

mscheltienne:
I wonder if it has to do with the caching.. I'll have a look, but not today. I've spent enough time on CIs for one day ^^

adam2392:
Kay fixed it. Merging once CIs green.

mscheltienne:
Don't merge yet, I still have a couple of points I want to check here ;)

mscheltienne:
My bad for the `checkout`, as it was on a single line without name/`run` command, and for some reason in a different color on my IDE, I did not assume it was an action 😅 

-----------------

Looking a bit more at the `config.yml` for CircleCI and at their documentation:

```
- restore_cache:
            keys:
              - source-cache-iclabel
        - checkout
        - run:
            name: Complete checkout
            command: |
              if ! git remote -v | grep upstream; then
                git remote add upstream https://github.com/mne-tools/mne-icalabel.git
              fi
              git remote set-url upstream https://github.com/mne-tools/mne-icalabel.git
              git fetch upstream
        - save_cache:
            key: source-cache-iclabel
            paths:
              - ".git"
```

Why is the `Complete checkout` step required? Looking at [Circle's documentation](https://circleci.com/docs/configuration-reference#checkout), `checkout` will clone the repository if the path (working directory) does not exist and fetch the origin if the path is a git repository (i.e. if the cache is restored).

-----------------

I moved the `Check-skip` as the first step, but indeed it can not work if the repo is not checkout first. My bad again 😅 

-----------------

```
- add_ssh_keys:
            fingerprints:
              - "d6:88:6b:a0:80:bf:14:8e:80:2e:ed:11:52:26:37:68"
```

Any idea what this is used for?

-----------------

And last point, for Qt, you installed `qt5-defaults` in the `Set BASH_ENV` step while you added `PySide6` to the requirements for the GUI. I think we do not want to add any of the 4 possible Qt bindings to the GUI extra-keyword. If we look at `mne-qt-browser` and its requirements https://github.com/mne-tools/mne-qt-browser/blob/main/requirements.txt you can see that `qtpy` is present but none on the Qt bindings. It is up to the user to install the Qt bindings of their choice, while the package can raise an explicit error prompting the user to install one of the 4 bindings if it is absent. 

This is why I have set `PyQt5` in the `test` extra-requirements, to make our life simpler and test only on `PyQt5`. If you look at the MNE QT Browser again, the test requirements do not include any Qt bindings because the tests run on all 4: https://github.com/mne-tools/mne-qt-browser/blob/2630213caef5a73fc2453fb816c00d1222442591/.github/workflows/qt_viz_tests.yml#L130-L134

@adam2392 Proposition is to:
- Remove PySide6 from the `gui` requirements
- Add PyQt5 to the `doc` requirements (to use the same binding on both tests and doc)
- Use the `_check_qt_version` function you added to raise an explicit message if the GUI can not find any Qt bindings.
- In a future PR, maybe, let's test only the GUI tests on all 4 Qt bindings.

-----------------

I'm gonna try to cut some part out, let's see what happens.

mscheltienne:
For the first bullet point `Complete checkout`, I guess it's because we only cache the `.git` folder and the actual files have to be fetch?

mscheltienne:
@adam2392 Couple of additional points to look at below.

### Circle

This path does not exist https://github.com/mscheltienne/mne-icalabel/blob/cc372ad9932c2086c5d946515ece07fee82c20be/.circleci/config.yml#L130-L132

The `linkcheck` workflow is always skipped https://github.com/mscheltienne/mne-icalabel/blob/cc372ad9932c2086c5d946515ece07fee82c20be/.circleci/config.yml#L139-L158

### GitHub

Where is it downloading artifacts from? https://github.com/mscheltienne/mne-icalabel/blob/cc372ad9932c2086c5d946515ece07fee82c20be/.github/workflows/unit_tests.yml#L186-L190

adam2392:
> ### Circle
> This path does not exist https://github.com/mscheltienne/mne-icalabel/blob/cc372ad9932c2086c5d946515ece07fee82c20be/.circleci/config.yml#L130-L132

Added logic to use both directories differently for docs-deployment to gh-pages. Not 100% sure it's working "properly", but I think it is...

> The `linkcheck` workflow is always skipped https://github.com/mscheltienne/mne-icalabel/blob/cc372ad9932c2086c5d946515ece07fee82c20be/.circleci/config.yml#L139-L158

Ah yes, I added in a weekly check, that MNE-Python has in their config file, which will now run it.

> ### GitHub
> Where is it downloading artifacts from? https://github.com/mscheltienne/mne-icalabel/blob/cc372ad9932c2086c5d946515ece07fee82c20be/.github/workflows/unit_tests.yml#L186-L190

I added the upload-artifacts for the build step, so now it will store those artifacts to be used when we slate a release on GH.

^ So to make a release:

1. Make new PR that updates version to `0.3` without the `dev0`, adds summary text to the `whats_new.rst` 
2. Build docs manually
3. Upload docs to `gh-pages` branch for new `stable` directory
4. Move `stable` directory to its version e.g. `v0.2` if the new version that is released is `v0.3`.
5. Merge PR
6. Make a release on GH
7. GH-Actions should automatically upload that release to pypi
8. conda-forge recipe should automatically upload new conda-forge recipe
9. Make new PR to modify the `whats_new.rst` and move contents to `whats_new_previous.rst`
10. Modify version to `v0.4dev0` if say previously released version was `v0.3`, and then merge PR

adam2392:
I have also added the CITATION.cff file and added some TODO:

- to add some nicety to our releases, it would be great for the script to be refactored to pull from `whats_new.rst` and added into the release notes. Note this gh-action was mainly modified from the https://github.com/allenai/python-package-template

adam2392:
Seems like codecov is messing up... but can't tell if this is the issue? https://github.com/codecov/codecov-action/issues/598

adam2392:
If the rest works, I'll try explicitly setting the version for codecov.

mscheltienne:
I tried running it on a different job, just to see if the timeout occurs again.

mscheltienne:
@adam2393 Which commands from the `Makefile` are you using? I'd like to get rid of what we don't use.. as it is, I'm not sure even half of them are working anyway.

mscheltienne:
And I think that line: https://github.com/mscheltienne/mne-icalabel/blob/9d35476815d74fd941545855d65d0f59c87741ad/.circleci/config.yml#L253
and that block: https://github.com/mscheltienne/mne-icalabel/blob/9d35476815d74fd941545855d65d0f59c87741ad/.circleci/config.yml#L5-L8
Can be both removed, the schedule parameters is not used in this workflow.

adam2392:
> @adam2393 Which commands from the `Makefile` are you using? I'd like to get rid of what we don't use.. as it is, I'm not sure even half of them are working anyway.

The ones for running checks, building docs, py testing and "clean". Everything else I don't use

adam2392:
> And I think that line: https://github.com/mscheltienne/mne-icalabel/blob/9d35476815d74fd941545855d65d0f59c87741ad/.circleci/config.yml#L253
> and that block: https://github.com/mscheltienne/mne-icalabel/blob/9d35476815d74fd941545855d65d0f59c87741ad/.circleci/config.yml#L5-L8
> Can be both removed, the schedule parameters is not used in this workflow.

Done

I've also cleaned up the Makefile now and edited the installation page.

adam2392:
> I tried running it on a different job, just to see if the timeout occurs again.

Python 3.9 isn't used, so codecov doesn't upload tho I just realized.

mscheltienne:
Ahhh yes, that's a dumb one. I forgot that it tests only on the min/max version. Let's revert to 3.10 and see if the issue persists.

adam2392:
Kay it works! one time error...

adam2392:
In a downstream PR, we should try to do some maintenance to bring up test coverage to ~95%. Most likely these will come as we discover bugs w the added code for BIDs/GUI/etc. :p

mscheltienne:
Timeouts can happen. Let's reread a bit through these changes both of us, and see if we have anything else we want to change. And then, good to merge.

.github/workflows/unit_tests.yml
```diff
@@ -1,217 +1,212 @@
...
+          python -m pip install --progress-bar off --upgrade pip setuptools wheel
+          python -m pip install --progress-bar off .[style]
+      - name: Setup flake8 annotations
+        uses: rbialon/flake8-annotations@v1
+      - name: Run flake8
+        uses: py-actions/flake8@v2
+        with:
+          path: "mne_icalabel"
+      - name: Run isort
+        uses: jamescurtin/isort-action@master
```
adam2392, line 38: For these, I'm not 100% knowledgeable, is this the official version? E.g. I found this on the web: https://github.com/isort/isort-action.

Same for `flake8` and `mypy`?

adam2392 (commented):
Everything else LGTM.

.github/workflows/unit_tests.yml
mscheltienne: I'm guessing `jamescurtin/isort-action` got renamed/moved to the `isort` namespace.. as I can't find anymore the GitHub page of `jamescurtin/isort-action` and as he his the main (almost only) contributor to that action ;)

Also: https://pycqa.github.io/isort/docs/configuration/github_action.html
Which does point to the new link, but I'm 99% sure it was pointing to jamescurtin namespace before ;)

I'll check them and update them.

mscheltienne:
LGTM, merging if CIs come back green.

mscheltienne:
@adam2392 Looks like the doc-deploy image does not exist.

PR #89, opened 2022-08-04
sidecar .json file saved separately

anandsaini024:
Used suffix 'digitizer' to save it separately from the other components label .tsv file in mne_iclabel/annotations/mara.py

### DIFF
--- a/mne_icalabel/annotation/bids.py
+++ b/mne_icalabel/annotation/bids.py
@@ -82,7 +82,7 @@ def write_components_tsv(ica: ICA, fname):
         "'muscle artifact', 'eye blink', 'heart beat', 'line noise', "
         "'channel noise', 'other']",
     }
-    fname = fname.copy().update(extension=".json")
+    fname = fname.copy().update(suffix="digitizer", extension=".json")
     if not fname.fpath.exists():
         _write_json(fname, component_json)
     else:


adam2392:
Sorry I'm not 100% following here. What is the goal of this change @anandsaini024 ?

anandsaini024:
> Sorry I'm not 100% following here. What is the goal of this change @anandsaini024 ?

The file mara.py in [improve_icalabel](https://github.com/adam2392/improve_icalabel)/[mnestudy](https://github.com/adam2392/improve_icalabel/tree/master/mnestudy)/[ica_to_bids](https://github.com/adam2392/improve_icalabel/tree/master/mnestudy/ica_to_bids)/ was not writing all the files in the output directory. There were minor errors and those are corrected. 

Although, It still does not give the desired output as the number of component column is repeating in each iteration https://github.com/mne-tools/mne-icalabel/blob/f740a593bdcbb1c0d390216270be6af2faf522fc/mne_icalabel/annotation/bids.py#L163. You can run the mara.py to check the output. it does not through an error now.

mscheltienne:
I'll have to check.. but this change does not make sense to me either :/

anandsaini024:
> I'll have to check.. but this change does not make sense to me either :/

This change is to generate a sidecar in .json format apart from the .tsv file. Earlier this was not able to generate a separate .json file because https://github.com/mne-tools/mne-icalabel/blob/f740a593bdcbb1c0d390216270be6af2faf522fc/mne_icalabel/annotation/bids.py#L86-L89 update_sidecar_json only works if it's a json file. Basically there was only one file which is .tsv before this change (atleast on my computer). Please check and let me know if it was unnecessary.

PR #96, opened 2022-08-09
[DOC] Remove typing-extensions and replace outdated gh-substitutions by sphinx-issues

mscheltienne:
@adam2392 This is a proposal with 2 parts:
- Remove `typing-extensions` from the documentation requirements: we don't use it.
- Replace MNE custom sphinx-extension `gh-substitutions` with `sphinx-issues`. I can revert this one if you prefer.

MNE uses the `:gh:` role to points to GitHub. `sphinx-issues` does the same thing, but with the `:issue:` and `:pr:` role(and can support other issues tracker). It also adds the `:commit:` role. Overall, it just helps us get rid of some files. I suspect the `gh-substitutions` extension was added a long time ago ;)

Downside: it sets `mne-icalabel` changelog apart from the rest of the MNE ecosystem.

Let me know what you think about this change.


adam2392:
I'm okay with modernizing that. Scikit-learn uses `sphinx_issues` too.

adam2392 (approved):

PR #108, opened 2022-08-19
Improve description of supported datatypes by ICLabel

### LINKED ISSUE #105: [DOC] Clarify the types of data to be used
I am unfamiliar with the original ICLabel tool, and in reading the docs and subsequently applying mne-icalabel it to my own MEG data I got confused for which types of data it is intended to be used. The paper and introduction in the documention mention "Scalp electroencephalography (EEG) and magnetoencephalography (MEG)" explicitly, and summarize generally "mne-icalabel is a Python package for labeling independent components that stem from an [Independent Component Analysis (ICA)](https://en.wikipedia.org/wiki/Independent_component_analysis)." This made me think that I could use the tool on ICAs fitted on non-EEG data such as MEG or ieeg or ECoG as well. The warning emitted by `label_components()` (also mentioned in #86) however somewhat implies I would need to have EEG data:

```
<ipython-input-7-1dec4ffcd5a6>:1: RuntimeWarning: The provided Raw instance does not seem to be referenced to a common average reference (CAR). ICLabel was designed to classify features extracted from an EEG dataset referenced to a CAR (see the 'set_eeg_reference()' method for Raw and Epochs instances).
```
and ndeed I can not get the examples to run on my own data without EEG channels. 
The original ICLabel [paper's abstract](https://www.sciencedirect.com/science/article/abs/pii/S1053811919304185?via%3Dihub) mentions EEG data only. A comment in the [tutorial](https://mne.tools/mne-icalabel/dev/auto_examples/iclabel_automatic_artifact_correction_ica.html#example-eog-and-ecg-artifact-repair) seems to suggest it is even more restricted:

> \# Note: for this example, we are using ICLabel which has only
\# been validated and works for EEG systems with less than 32 electrodes.

In response to #86 , PR #87  also introduced the following changes in the tutorial to get rid of a few warnings:

```
ica = ICA(n_components=15, max_iter="auto", random_state=97)
# Before fitting ICA, we will apply a common average referencing.
filt_raw = filt_raw.set_eeg_reference("average")

# Note: we will use the 'infomax' method for fitting the ICA because
# that is what is supported in the ICLabel model. In practice, one can
# use any method they choose.
ica = ICA(
    n_components=15,
    max_iter="auto",
    method="infomax",
    random_state=97,
    fit_params=dict(extended=True),
)
```
I think it would be useful if you could clarify the requirements mne-icalabel has for ICA data, and state more prominently which data is suitable and which data isn't suitable  to help people unfamiliar with the original EEGLab ICLabel tool gauge quickly if it is useful for them. Likewise a clarification about the requirements of common average referencing and non-fastica ICA, probably even in the docstring of the function,  makes sense if it is a strong requirement or recommendation, which isn't immediately obvious to me from code and documentation. The code emitting the warning reads as if bad things happen if a non-infomax method is used:

https://github.com/mne-tools/mne-icalabel/blob/b91ac87068d3e43076298574b54d98e1633da4f1/mne_icalabel/iclabel/features.py#L61-L73

But the comment in the tutorial says "In practice, one can use any method they choose." If non-infomax methods are not a problem, why the warning? If non-infomax methods are not benchmarked/tested/validated, I think its worth an easily findable note on that in the docstring so that anyone can learn about this without reading the walk-throughs in detail.

mscheltienne:
Closes #105

### DIFF
--- a/doc/api.rst
+++ b/doc/api.rst
@@ -28,7 +28,10 @@ The model was ported from matconvnet using `pytorch <https://pytorch.org/>`_.
 
 ICLabel is designed to classify ICs fitted with an extended infomax ICA
 decomposition algorithm on EEG datasets referenced to a common average and
-filtered between [1., 100.] Hz.
+filtered between [1., 100.] Hz. It is possible to run ICLabel on datasets that
+do not meet those specification, but the classification performance
+might be negatively impacted. Moreover, the ICLabel paper did not study the
+effects of these preprocessing steps.
 
 Architecture:
 
--- a/doc/api.rst
+++ b/doc/api.rst
@@ -41,10 +41,9 @@ Architecture:
    :align: center
 
 The model has three inputs: image (topomap), psd, and autocorrelation features.
-To encourage generalization, the image features are rotated and negated, thus
-quadrupling the feature. The psd and autocorrelation features are copied to the
-new image features. Then, the predicted probabilities are averaged over all four
-images.
+To encourage generalization, the image feature is rotated and negated, thus
+quadrupling the feature. After 3 convolutional layer with a ReLu activation,
+the 3 features are concatenated for the final layer.
 
 .. currentmodule:: mne_icalabel.iclabel
 
--- a/doc/api.rst
+++ b/doc/api.rst
@@ -26,6 +26,9 @@ ICLabel
 This is the model originally available for `EEGLab <https://github.com/sccn/ICLabel>`_.
 The model was ported from matconvnet using `pytorch <https://pytorch.org/>`_.
 
+ICLabel is designed to classify ICs fitted with an extended infomax ICA decomposition
+algorithm on EEG datasets referenced to a common average.
+
 Architecture:
 
 .. image:: _static/ICLabel_DagNN_Architecture.png
@@ -41,6 +44,7 @@ are copied to the new image features. Then, the predicted probabilities are aver
 .. autosummary::
    :toctree: generated/
 
+   iclabel_label_components
    get_iclabel_features
    run_iclabel
 
--- a/doc/api.rst
+++ b/doc/api.rst
@@ -26,18 +26,22 @@ ICLabel
 This is the model originally available for `EEGLab <https://github.com/sccn/ICLabel>`_.
 The model was ported from matconvnet using `pytorch <https://pytorch.org/>`_.
 
-ICLabel is designed to classify ICs fitted with an extended infomax ICA decomposition
-algorithm on EEG datasets referenced to a common average.
+ICLabel is designed to classify ICs fitted with an extended infomax ICA
+decomposition algorithm on EEG datasets referenced to a common average and
+filtered between [1., 100.] Hz.
 
 Architecture:
 
 .. image:: _static/ICLabel_DagNN_Architecture.png
    :width: 400
    :alt: ICLabel Neural Network Architecture
+   :align: center
 
-The model has three inputs: image, psd, and autocorrelation features. To encourage generalization, the image
-features are rotated and negated, thus quadrupling the feature. The psd and autocorrelation features
-are copied to the new image features. Then, the predicted probabilities are averaged over all four images.
+The model has three inputs: image (topomap), psd, and autocorrelation features.
+To encourage generalization, the image features are rotated and negated, thus
+quadrupling the feature. The psd and autocorrelation features are copied to the
+new image features. Then, the predicted probabilities are averaged over all four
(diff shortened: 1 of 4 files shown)


mscheltienne:
@adam2392 "Check the rendered docs here!" does not seem to work.

image

mscheltienne:
Renders:

- API page: https://output.circle-artifacts.com/output/job/8a02721f-5bd8-419e-acda-75b912e2800f/artifacts/0/stable/api.html

image
image

- `iclabel_label_components`: https://output.circle-artifacts.com/output/job/8a02721f-5bd8-419e-acda-75b912e2800f/artifacts/0/stable/generated/mne_icalabel.iclabel.iclabel_label_components.html#mne_icalabel.iclabel.iclabel_label_components

- Tutorial: https://output.circle-artifacts.com/output/job/8a02721f-5bd8-419e-acda-75b912e2800f/artifacts/0/stable/auto_examples/iclabel_automatic_artifact_correction_ica.html#sphx-glr-auto-examples-iclabel-automatic-artifact-correction-ica-py

mscheltienne:
In the tutorial:

image

That's actually a pretty bad fit, no? The first IC with the blink looks super noisy.

adam2392:
> That's actually a pretty bad fit, no? The first IC with the blink looks super noisy.

Fair point :p, but I guess I meant that the fit captures the eye "blinks" but it's just reverse amplitude.

doc/api.rst
```diff
@@ -26,21 +26,29 @@ ICLabel
 This is the model originally available for `EEGLab <https://github.com/sccn/ICLabel>`_.
 The model was ported from matconvnet using `pytorch <https://pytorch.org/>`_.
 
+ICLabel is designed to classify ICs fitted with an extended infomax ICA
+decomposition algorithm on EEG datasets referenced to a common average and
+filtered between [1., 100.] Hz.
```
adam2392, line 31: Perhaps we can say also something like:

We allow users to still run their EEG data if they don't meet these specifications, but we do not make any guarantees for software performance. Moreover, the ICLabel paper did not study the effects of these preprocessing steps.

doc/api.rst
```diff
@@ -26,21 +26,29 @@ ICLabel
...
+   :align: center
 
-The model has three inputs: image, psd, and autocorrelation features. To encourage generalization, the image
-features are rotated and negated, thus quadrupling the feature. The psd and autocorrelation features
-are copied to the new image features. Then, the predicted probabilities are averaged over all four images.
+The model has three inputs: image (topomap), psd, and autocorrelation features.
+To encourage generalization, the image features are rotated and negated, thus
+quadrupling the feature. The psd and autocorrelation features are copied to the
+new image features. Then, the predicted probabilities are averaged over all four
+images.
```
adam2392, line 44: four or three images?

adam2392 (approved):
LGTM mostly

The style check should be fixed once you merge main branch changes.

doc/api.rst
mscheltienne: Have a look at how I rephrase it, I don't really get what you meant by this sentence anyway 😅

adam2392: Oh haha I meant. Aren't the features comprised of 3 images? wdym by four image?

mscheltienne: The comment was clear, it's the sentence in the doc that is not 😅 
The PSD and autocorrelation are 1D: https://github.com/mne-tools/mne-icalabel/blob/0ae320a7c90cf7357547a34b4abf663511a1e0ad/mne_icalabel/iclabel/network.py#L52-L58 
Sure it's a Conv2D, but with a kernel size `(1, 3)`. So basically a Conv1D. Thus the terminology of `image` is already misleading.

mscheltienne: Or am I missing something?

adam2392: Ah I see. Yeah if I wrote this then I'm not sure what I be meant :p. 

Perhaps something like 3 features?

mscheltienne: That's my assumption as well :p 
Thus the rework as: 

> The model has three inputs: image (topomap), psd, and autocorrelation features.
> To encourage generalization, the image feature is rotated and negated, thus
> quadrupling the feature. After 3 convolutional layer with a ReLu activation,
> the 3 features are concatenated for the final layer.

Ok for you or would you like to change something else here?

adam2392: Yeah that looks good to me and more clear!

adam2392 (approved):

adam2392:
Thanks @mscheltienne !

PR #115, opened 2022-08-27
Fix codecov

adam2392:
PR Description
--------------

Merge checklist
---------------

- [ ] All comments resolved
- [ ] This is not your own PR
- [ ] All CIs are happy
- [ ] New contributors have been added to [CITATION.cff](https://github.com/mne-tools/mne-icalabel/blob/main/CITATION.cff)

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,0 +1,5 @@
+fixes:
+  - "main/::"
+
+github_checks:
+  annotations: false
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,5 +1,0 @@
-fixes:
-  - "main/::"
-
-github_checks:
-  annotations: false
--- a/.github/workflows/draft-pdf.yml
+++ b/.github/workflows/draft-pdf.yml
@@ -2,7 +2,7 @@ on:
   push:
     branches: [main]
     paths:
-      - 'paper/*'
+      - "paper/*"
 
 jobs:
   paper:
--- a/.github/workflows/unit_tests.yml
+++ b/.github/workflows/unit_tests.yml
@@ -159,15 +159,16 @@ jobs:
       - name: Run pytest  # headless via Xvfb on linux
         uses: GabrielBB/xvfb-action@v1.6
         with:
-          run: pytest main/mne_icalabel --cov=main/mne_icalabel --cov-report=xml --cov-config=main/pyproject.toml
+          run: pytest main/mne_icalabel --cov-config=main/pyproject.toml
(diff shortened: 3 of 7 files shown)


mscheltienne:
I've been playing with `pytest`, `pytest-cov` and `codecov` for an hour on 2 different projects. Both other projects work like a charm, so I copy-pasted the exact configuration. 🤞

mscheltienne:
I was still not happy with the coverage configuration.

Now it should ignore the folders in `omits` correctly, no matter where the `pytest` command is run from. The key point was to remove the `source` definition from the configuration, because it was fine for CIs, but not for local run since `pytest package --cov=package` would not overwrite the source set in the configuration.

adam2392:
Why aren't the rest of the GH actions running? Is there a syntax error?

mscheltienne:
https://github.com/mne-tools/mne-icalabel/blob/2a201830e46c5a021dd7d9f0c13de10887c68a8f/.github/workflows/unit_tests.yml#L6-L8

mscheltienne:
We can remove the condition on paths if you want

adam2392:
Oh nah that sounds good. Altho... if we want to say change the `pyproject.toml` shouldn't we run at least the `build` job?

mscheltienne:
Yes, but you can't do that so easily if you have only one file with all the tests inside. You have to check the GitHub events in an `if` condition instead of relying on workflow triggers.

What I usually do is that I have separate files for build, doc-build, tests, style, ... see for instance this sandbox repository: https://github.com/mscheltienne/simple-stimuli/tree/main/.github/workflows

With the build workflow: https://github.com/mscheltienne/simple-stimuli/blob/c2282ae5438d82a6ec88cf6c707f5e9fb6b525f0/.github/workflows/build.yml#L5-L9

--------------

Good points: 
- Smaller workflow files, more readable, with fewer mistakes
- Better management of workflow triggers
- One badge for each workflow

Downsides:
- You can't use 'depends on' between workflows (files), only between jobs. 

--------------

Now, to get back to the trigger of `pytest` only on changes to `.py` files, it sounds good as indeed it will not fail if there has been no changes to those files; but in practice, it's a bit of a pain because if you want to change the workflow in a PR, it does not execute it. So better for day-to-day development by contributors, but worst for maintenance of the workflow itself.

adam2392:
Good to merge?

mscheltienne (approved):

mscheltienne:
I triggered the workflow manually, upload to codecov and `omit` worked:

image

PR #120, opened 2022-12-02
[MNT] Drop support for Python 3.7 and add support for Python 3.11

mscheltienne:
With the new version of Python, we should drop support for the older version and add support for the latest.
It's probably not going to be green for now.


adam2392:
Looks like not all CI running. Do we need to do it manually?

mscheltienne:
Not sure why, I'll try to check later.

mscheltienne:
The 2 first issues I see:
- `qt5-defaults` is not available anymore in Ubuntu 22.04 LTS
- PyTorch did not release yet the wheels for Python 3.11: https://github.com/pytorch/pytorch/issues/86566

mne_icalabel/gui/_label_components.py
```diff
@@ -1,7 +1,7 @@
 from typing import Dict, List, Union
 
 from matplotlib import pyplot as plt
-from matplotlib.backends.backend_qt5agg import FigureCanvasQTAgg
+from matplotlib.backends.backend_qtagg import FigureCanvasQTAgg
```
mscheltienne, line 4: And that is why we should test with all 4 Qt bindings like the Qt-browser 😓

mscheltienne:
We are also missing the wheels for `h5py`: https://github.com/h5py/h5py/issues/2146

adam2392:
Just following up, so the main issue is `h5py` and  `torch` for Python3.11?

mne_icalabel/gui/_label_components.py
adam2392: What happened?

mscheltienne: IIRC, it was failing on the Linux CI when I changed the installation from 'qt5-defaut' to 'qt6-base-dev'.
I'm guessing the backend 'qt_agg' is compatible with all bindings, while 'qt5_agg' works only with Qt5.

mscheltienne:
Yes, h5py new release should be out soon.

adam2392:
Looks like pytorch tho also has a delay for 3.11. Prolly by Feb I'm guessing should be released... hopefully :p

mscheltienne:
Quick note, we'll also need to update the conda feedstock when we drop 3.7.

pyproject.toml
```diff
@@ -41,8 +41,7 @@ dependencies = [
     'scipy >= 1.2.0',
     'mne >= 1.1',
     'pooch',
-    'torch',
-    'importlib-metadata; python_version<"3.8"',
+    'torch >= 2.0',
```
mscheltienne, line 44: Great they finally released!
I don't think we need the (very restrictive) version pin however. A Python 3.11 install will use the 2.0 version while an older Python version will use whatever version (the most recent one with an available wheel).

```suggestion
    'torch',
```

PR #121, opened 2023-01-22
[MNT] Fix intersphinx for numpy/scipy

mscheltienne:
The numpy link to the `dev` doc is not valid anymore. 
I replaced both links for numpy and scipy with the current valid ones for the `stable` documentation.


mscheltienne:
Merging as the linux CIs will be fixed by #120

Issue #125, opened 2023-03-07
ONNX/Tensorflow Port


jacobf18:
I just wanted to inform everybody that I was working on porting the PyTorch code to work with other deep-learning frameworks. I am doing this because PyTorch doesn't work well with other python packages for bundling into standalone programs. Also, PyTorch is large compared to other packages like ONNX when all we are doing is inference. The changes are minor and restricted to the `network.py` file.

We could make it an option for which framework to use if people have trouble installing PyTorch or want to run this package on a system that cannot run PyTorch.

mscheltienne:
That sounds like a great idea! I don't know ONNX, but I do like Tensorflow and Keras a lot :)
That might also make packaging simpler (or not?) especially when it comes to `conda-forge`.

jacobf18:
ONNX is universal so it allows for porting to tensorflow/keras as well. From my initial tests, loading the ONNX runtime package is a lot faster than importing PyTorch.

jacobf18:
Got it working on my laptop. Will format the code and push here soon. Also, if we wanted we could host the model using git LFS so that it wasn’t inside the repo. (It’s only 11 MB so this might not be necessary.) ONNX models can be loaded anywhere similar to PyTorch models with a `onnx.load` command (https://github.com/onnx/models)

PR #127, opened 2023-03-20
Update utils: remove unused function

mscheltienne:
This PR removes `_check_qt_version` which was added with the GUI but not used in the end.
And it also add `import_optional_dependency` to handle optional dependencies, which I expect to be useful for #125. If not, we can remove it later as well.


adam2392:
Do we necessarily need `import_optional_dependency`? We can just nest the dependency into the function and it raises an import error if someone doesn't have it installed?

mscheltienne:
Not necessarily, it's just a proposal. 

The advantages are:
- Nicer error message when an optional dependency is missing
- One-line call instead of `try / except` blocks
I would still nest `import_optional_dependency` anyway.

I think there is an equivalent in `mne.utils` also, so we could just use this one too if we need.

PR #129, opened 2023-06-19
Added ONNX support

### LINKED ISSUE #125: ONNX/Tensorflow Port

jacobf18:
PR Description
--------------

Added ONNX support to iclabel. I changed the function `run_iclabel` to include a library parameter. This parameter's default value is set to use PyTorch. I added unit tests to check the network outputs against MATLAB. ONNX is faster to load and more lightweight than PyTorch.

The options for the `library` parameter are `pytorch` and `onnx` so far. It throws a `ValueError` for any other input.

Closes #125.

Merge checklist
---------------

- [ ] All comments resolved
- [ ] This is not your own PR
- [ ] All CIs are happy
- [ ] New contributors have been added to [CITATION.cff](https://github.com/mne-tools/mne-icalabel/blob/main/CITATION.cff)

### DIFF
--- a/mne_icalabel/iclabel/tests/test_network.py
+++ b/mne_icalabel/iclabel/tests/test_network.py
@@ -4,6 +4,7 @@ except ImportError:
     from importlib_resources import files  # type: ignore
 
 import numpy as np
+import onnxruntime as ort
 import pytest
 import torch
 from scipy.io import loadmat
@@ -16,6 +17,7 @@ dataset_path = icalabel.data_path() / "iclabel"
 
 # Network weights
 torch_iclabel_path = str(files("mne_icalabel.iclabel").joinpath("assets/iclabelNet.pt"))
+onnx_network_file = files("mne_icalabel.iclabel").joinpath("assets/ICLabelNet.onnx")
 matconvnet_iclabel_path = dataset_path / "network/netICL.mat"
 
 # Network forward pass input/output
@@ -35,7 +37,7 @@ iclabel_output_raw_path = dataset_path / "iclabel-output-raw.mat"
 iclabel_output_epo_path = dataset_path / "iclabel-output-epo.mat"
 
 
-def test_weights():
+def test_weights_pytorch():
     """Compare the weights of pytorch model and matconvnet model."""
     network_python = torch.load(torch_iclabel_path)
     network_matlab = loadmat(matconvnet_iclabel_path)
@@ -81,9 +83,9 @@ def test_weights():
         assert np.allclose(network_python[python_layer], weights_matlab[idx])
 
 
-def test_network_outputs():
+def test_network_outputs_pytorch():
     """
-    Compare that the ICLabel network in python and matlab outputs the same
+    Compare that the ICLabel network in pytorch and matlab outputs the same
     values for a common set of features (input to the forward pass).
 
     Notes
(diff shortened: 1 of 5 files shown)


mne_icalabel/iclabel/tests/test_network.py
```diff
@@ -126,14 +128,58 @@ def test_network_outputs():
...
+
 @pytest.mark.parametrize(
     "eeglab_feature_file, eeglab_feature_formatted_file",
     [
         (features_raw_path, features_formatted_raw_path),
         (features_epo_path, features_formatted_epo_path),
     ],
 )
-def test_format_input(eeglab_feature_file, eeglab_feature_formatted_file):
+def test_format_input_pytorch(eeglab_feature_file, eeglab_feature_formatted_file):
```
mscheltienne, line 182:
```suggestion
def test_format_input(eeglab_feature_file, eeglab_feature_formatted_file):
```

This is testing a function common to both. I would keep the old name.

mne_icalabel/iclabel/tests/test_network.py
```diff
@@ -126,14 +128,58 @@ def test_network_outputs():
...
+    # - elements in position [0, 2, 4] -> 1-element array with the var name
+    # - elements in position [1, 3, 5] -> data arrays
+    assert "in_image" == features[0][0]
+    images = features[1]
+    assert "in_psdmed" == features[2][0]
+    psd = features[3]
+    assert "in_autocorr" == features[4][0]
+    autocorr = features[5]
+
+    # reshape the features to fit torch format
```
mscheltienne, line 153:
```suggestion
    # reshape the features to fit onnx format
```

pyproject.toml
```diff
@@ -56,7 +57,7 @@ doc = [
     'memory_profiler',
     'numpydoc',
     'pydata-sphinx-theme',
-    'PyQt5',
+    'PyQt6',
```
mscheltienne, line 60: Any reason to change this?

mscheltienne (approved):
Thanks @jacobf18 ! That looks good to me.

As you said, we are going to need to add some logic to load the libraries. I'm also +1 to make Qt an optional dependency. Overall, +1 to merge as is and clean-up, add library selection logic, and so on.. in a second PR.

pyproject.toml
```diff
@@ -42,6 +42,7 @@ dependencies = [
     'mne >= 1.1',
     'pooch',
     'torch',
+    'onnxruntime',
```
adam2392, line 45: Is there a way we can make both optional dependencies, since the user only needs to install one?

adam2392 (commented):
If we can make the dependencies optional, then we should nest the import statements for torch and onnx inside the functions that use them. 

Perhaps we should add a helper function that does the import and checks that at least one of them is installed?

mscheltienne:
Yes, we need to add some selection logic and get rid of the import at the top-level. Probably requires also 2 separate files, one for `pytorch` and one for `onnx` since the classes defined inherit from `torch.nn` directly. Maybe some logic like `psychopy` has [here](https://github.com/psychopy/psychopy/blob/81dc09ae935515ea43d97fc5eef281457d3a8c83/psychopy/parallel/__init__.py#L34-L57) in an `__init__`.
Anyway, I think that could be done in a second PR which restructure the `iclabel` module before a 0.5 release.

adam2392:
Alright. Sounds good to me. LGTM with the exception of CI. Does it just need a restart possibly?

Thanks @jacobf18!

mscheltienne:
No the ubuntu ones have been broken for a while because of some xvfb stuff, it will be a good occasion to fix them as well..

mscheltienne:
@jacobf18 Do you want to give a shot to a logic to set `pytorch` and `onnx` as optional?
Ideally:
- `import mne_icalabel` and `from mne_icalabel import label_components` would work regardless of the installation status of the backends.
- `import mne_icalabel.iclabel` and `from mne_icalabel.iclabel import label_components` would fail if neither backend is available.

A couple of ideas: we should move the network part to a `mne_icalabel.iclabel.network` module, with 3 files: 
- `__init__.py` which attempts to load both in a similar fashion to psychopy with the drivers for the parallel port
- `pytorch.py` with the pytorch backend
- `onnx.py` with the onnx backend

Or maybe an empty `__init__.py` and some helper function which handles the loading of the backends.
The tests might also require a decorator `@requires_onnx` or `@requires_torch`.

-------

Let me know if you want to give it a shot, else I'll do it in the coming days. Once it's sorted out, we can release a 0.5 version with this addition, and maybe get this correctly packaged on `conda-forge` for windows.

mscheltienne:
Merging and picking this up today with another PR to add the selection of the backend. Let's not stall this :)

PR #130, opened 2023-06-26
Add support for the selection of onnx, torch without requiring both installed

mscheltienne:
It's still a draft.

IMO, we should ship one of the backend via pip install without requiring an optional key.
+1 to keep torch as "default" since that's what we started with. For now in the PR, I moved both dep. to an extra-key but I don't like this solution as it means that `pip install mne_icalabel` now leads to a "broken" package. 
However, we don't need to have the same default on PyPI and conda-forge, or the same default on all platform if that can make packaging on conda-forge easier.

This PR moves the network part to a separate module. The selection itself is donoe in `network/__init__.py` in `run_iclabel`. Seems to work locally, but I haven't run the full tests.

Also, @adam2392 would you mind changing the black default to 88 characters to match MNE?


mscheltienne:
And for now in this PR, selection can not be done at the top level, `mne_icalabel.label_components` wil default to the first backend found in the order `torch`, `onnx`. Selection can be done at the next level, `mne_icalabel.iclabel.iclabel_label_components` with the backend argument. I was thinking of adding a `set_backend` similar to `set_browser_backend` which would modify the behavior of `backend=None`.

mscheltienne:
> IMO, we should ship one of the backend via pip install without requiring an optional key.
> +1 to keep torch as "default" since that's what we started with. For now in the PR, I moved both dep. to an extra-key but I don't like this solution as it means that pip install mne_icalabel now leads to a "broken" package.

@adam2392 What is your opinion on this?

adam2392:
> > IMO, we should ship one of the backend via pip install without requiring an optional key.
> > +1 to keep torch as "default" since that's what we started with. For now in the PR, I moved both dep. to an extra-key but I don't like this solution as it means that pip install mne_icalabel now leads to a "broken" package.
> 
> @adam2392 What is your opinion on this?

I like the idea of this in principle in the long run because ideally we can run some other stuff that doesn't require the deep-learning models. But for now, if we really do "require" torch, I would say let's default to that.

adam2392 (approved):
LGTM once CIs work

mscheltienne:
OK, great. Let's go with that then, and let's try to setup ONNX as the default in the conda-forge package if that helps packaging on windows. CIs should come back green, merging once that's the case.

PR #136, opened 2023-08-23
MAINT: Compatiblity for MNE 1.6

mscheltienne:
Compatibility following https://github.com/mne-tools/mne-python/pull/11903
Should work with both MNE 1.5 and below and with MNE 1.6 and above.
It's probably worth to cut a release, or cut a release when MNE 1.6 rolls out.


adam2392 (approved):

adam2392:
LGTM  for 1.6

PR #193, opened 2024-07-25
FIX: explicitly set weights_only to avoid FutureWarning

### LINKED ISSUE #192: FutureWarning in latest pytorch release
I started seeing this in the CI's of another project I work on: https://github.com/lina-usc/pylossless/actions/runs/10085318831/job/27886113913

Looks like PyTorch 2.4 was just released today, and they throw a `FutureWarning` for an upcoming API change (see [changelog](https://github.com/pytorch/pytorch/releases/tag/v2.4.0) ):

> User warning when using torch.load with default weights_only=False value (https://github.com/pytorch/pytorch/pull/129239, https://github.com/pytorch/pytorch/pull/129396, https://github.com/pytorch/pytorch/pull/129509).
A warning is now raised if the weights_only value is not specified during a call to torch.load, encouraging users to
adopt the safest practice when loading weights.

It appears that the offending line is here:

 https://github.com/mne-tools/mne-icalabel/blob/1c36710dac5600942d976ade0c2ee43afef2e368/mne_icalabel/iclabel/network/torch.py#L201C1-L202C1

I can open up a PR in the morning unless someone else beats me to it.

scott-huberty:
Closes #192 

If I'm understanding the Torch 2.4 changelog correctly, you just need to explicitly pass a value to `weights_only` to suppress the warning. Since the default is already `False`, I am just explicitly setting it here, so this should be backward compatible

I'll double check that the CI installs 2.4 so we can make sure this fix works. 

- [ ] All comments resolved
- [ ] All CIs are happy
- [x] Updated a changelog entry and contributor name in [whats_new.rst](https://github.com/mne-tools/mne-icalabel/blob/main/doc/whats_new.rst) is updated
- [x] PR description includes phrase "closes <#issue-number>"

### DIFF
--- a/mne_icalabel/iclabel/network/tests/test_network.py
+++ b/mne_icalabel/iclabel/network/tests/test_network.py
@@ -36,8 +36,8 @@ iclabel_output_epo_path = dataset_path / "iclabel-output-epo.mat"
 
 def test_weights_pytorch():
     """Compare the weights of pytorch model and matconvnet model."""
-    network_python = torch.load(torch_iclabel_path)
-    network_matlab = loadmat(matconvnet_iclabel_path)
+    network_python = torch.load(torch_iclabel_path, weights_only=False)
+    network_matlab = loadmat(matconvnet_iclabel_path, weights_only=False)
 
     # load weights from matlab network
     weights_matlab = network_matlab["params"]["value"][0, :]
--- a/mne_icalabel/iclabel/network/tests/test_network.py
+++ b/mne_icalabel/iclabel/network/tests/test_network.py
@@ -119,7 +119,7 @@ def test_network_outputs_pytorch():
 
     # run the forward pass on pytorch
     iclabel_net = ICLabelNet()
-    iclabel_net.load_state_dict(torch.load(torch_iclabel_path))
+    iclabel_net.load_state_dict(torch.load(torch_iclabel_path, weights_only=False))
     torch_labels = iclabel_net(images, psd, autocorr)
     torch_labels = torch_labels.detach().numpy()  # (30, 7)
 
--- a/mne_icalabel/iclabel/network/tests/test_network.py
+++ b/mne_icalabel/iclabel/network/tests/test_network.py
@@ -37,7 +37,7 @@ iclabel_output_epo_path = dataset_path / "iclabel-output-epo.mat"
 def test_weights_pytorch():
     """Compare the weights of pytorch model and matconvnet model."""
     network_python = torch.load(torch_iclabel_path, weights_only=False)
-    network_matlab = loadmat(matconvnet_iclabel_path, weights_only=False)
+    network_matlab = loadmat(matconvnet_iclabel_path)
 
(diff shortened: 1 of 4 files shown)


mne_icalabel/iclabel/network/tests/test_network.py
```diff
@@ -36,8 +36,8 @@
 
 def test_weights_pytorch():
     """Compare the weights of pytorch model and matconvnet model."""
-    network_python = torch.load(torch_iclabel_path)
-    network_matlab = loadmat(matconvnet_iclabel_path)
+    network_python = torch.load(torch_iclabel_path, weights_only=False)
+    network_matlab = loadmat(matconvnet_iclabel_path, weights_only=False)
```
mscheltienne, line 40:
```suggestion
    network_matlab = loadmat(matconvnet_iclabel_path)
```

mscheltienne (approved):
Thanks @scott-huberty 
We are loading a file we distribute, thus I think it's safe to keep loading with `weights_only=False`. 
I pushed 2 commits which should get the CIs green.

adam2392:
https://pytorch.org/docs/stable/generated/torch.load.html

Seems to suggest if we set to True we introduce security issues. I don't see any issue with just keeping it as default of false.

scott-huberty:
> https://pytorch.org/docs/stable/generated/torch.load.html

I think it's the other way around. `weights_only=True` is safer because it restricts the deserialization to only tensors, primitive types, and dictionaries etc. whereas with `weights_only=False`, `pickle` could deserialize arbitrary/malicious python objects. 

But I think they made/kept the default `weights_only=False` for backwards compatibility reasons: 

https://github.com/pytorch/pytorch/pull/87443

adam2392:
Sorry yeah oops mixed it up. 

But yeah I don't see any reason we can't just use weights_only as True. Is there any issue with this? I don't think we load any non PyTorch primitives.

mscheltienne:
> But yeah I don't see any reason we can't just use weights_only as True. Is there any issue with this? I don't think we load any non PyTorch primitives.

Well that's the part I don't know and am not familiar with.. but the unit tests seem to agree since they come back green locally with `weights_only=True`.

scott-huberty:
Agreed that ideally `weights_only=True` is preferred/safer but I also wasn't sure if that would break your code so went with `False`. I'm OK with changing it to `True` if you all are.

mscheltienne:
Let's go with `True`, IIRC we have tests which compare a forward pass in the model against numerical value outputted by the same forward pas in MATLAB, they would have failed if `True` was messing up something.

Thanks @scott-huberty I'll cut a release tomorrow to get your CIs green.

scott-huberty:
> Thanks @scott-huberty I'll cut a release tomorrow to get your CIs green.

Thx @mscheltienne ! No rush in the meantime I'm just going make use of `filterwarnings` ; )

mscheltienne:
@scott-huberty 0.7 is out ;) Thanks!

PR #207, opened 2024-10-18
add MEGnet to make MNE-ICALabel work on MEG data

colehank:
PR Description
--------------
### MEGnet Integration

Hi everyone, 

I integrated **[MEGnet](https://github.com/DeepLearningForPrecisionHealthLab/MegNET_2020)**  into `mne_icalabel/megnet`. Here’s a summary of the main work:

1. **ONNX Conversion**: Transformed the original Keras model into ONNX format and confirmed that the prediction outputs remain consistent. 
   
2. **Feature Simulation**: Added `megnet/features.py` to simulate the **BrainStorm ICA topomap** used in MEGnet training (for cases where MNE's ICA topomap is used ).
   
3.  **Prediction Function Simplification**: Refined the prediction logic in `megnet/label_components.py` for simplicity. Also, the outputs now align with the current EEG `label_components`, using the keys:
   - `label`
   - `y_pred_proba`

You can run the following commands to check out the prediction performance of MEGnet on MNE’s sample data:

```bash
python megnet/features.py
python megnet/label_components.py
```
However, to successfully integrate into mne-icalabel, changes need to be made to the contents of `mne-icalabel/label_components`. I’m worried that, given my understanding of the project, I might mess it up, so all the work is only within mne_icalabel/megnet.

Merge checklist
---------------

- [ ] All comments resolved
- [ ] All CIs are happy
- [ ] Updated a changelog entry and contributor name in [whats_new.rst](https://github.com/mne-tools/mne-icalabel/blob/main/doc/whats_new.rst) is updated

### DIFF
--- a/mne_icalabel/megnet/_utils.py
+++ b/mne_icalabel/megnet/_utils.py
@@ -1,0 +1,45 @@
+#%%
+import numpy as np
+
+# Conversion functions
+def cart2sph(x, y, z):
+    xy = np.sqrt(x * x + y * y)
+    r = np.sqrt(x * x + y * y + z * z)
+    theta = np.arctan2(y, x)
+    phi = np.arctan2(z, xy)
+    return r, theta, phi
+
+def pol2cart(rho, phi):
+    x = rho * np.cos(phi)
+    y = rho * np.sin(phi)
+    return x, y
+
+
+def make_head_outlines(sphere, pos, outlines, clip_origin):
+    assert isinstance(sphere, np.ndarray)
+    x, y, _, radius = sphere
+    del sphere
+
+    ll = np.linspace(0, 2 * np.pi, 101)
+    head_x = np.cos(ll) * radius * 1.01 + x
+    head_y = np.sin(ll) * radius * 1.01 + y
+    dx = np.exp(np.arccos(np.deg2rad(12)) * 1j)
+    dx, dy = dx.real, dx.imag
+    
+    outlines_dict = dict(head=(head_x, head_y))
+    
+    mask_scale = 1.
+    mask_scale = max(
+        mask_scale, np.linalg.norm(pos, axis=1).max() * 1.01 / radius
+    )
+    
+    outlines_dict['mask_pos'] = (mask_scale * head_x, mask_scale * head_y)
+    clip_radius = radius * mask_scale
+    outlines_dict['clip_radius'] = (clip_radius,) * 2
+    outlines_dict['clip_origin'] = clip_origin
+    
+    outlines = outlines_dict
+    
+    return outlines
+
+
--- a/mne_icalabel/megnet/_utils.py
+++ b/mne_icalabel/megnet/_utils.py
@@ -1,5 +1,6 @@
-#%%
+# %%
 import numpy as np
+
 
 # Conversion functions
 def cart2sph(x, y, z):
@@ -8,6 +9,7 @@ def cart2sph(x, y, z):
     theta = np.arctan2(y, x)
     phi = np.arctan2(z, xy)
     return r, theta, phi
+
 
 def pol2cart(rho, phi):
     x = rho * np.cos(phi)
@@ -24,22 +26,19 @@ def make_head_outlines(sphere, pos, outlines, clip_origin):
     head_x = np.cos(ll) * radius * 1.01 + x
     head_y = np.sin(ll) * radius * 1.01 + y
     dx = np.exp(np.arccos(np.deg2rad(12)) * 1j)
-    dx, dy = dx.real, dx.imag
-    
+    dx, _ = dx.real, dx.imag
+
     outlines_dict = dict(head=(head_x, head_y))
-    
-    mask_scale = 1.
-    mask_scale = max(
-        mask_scale, np.linalg.norm(pos, axis=1).max() * 1.01 / radius
-    )
-    
-    outlines_dict['mask_pos'] = (mask_scale * head_x, mask_scale * head_y)
+
+    mask_scale = 1.0
+    max_norm = np.linalg.norm(pos, axis=1).max()
+    mask_scale = max(mask_scale, max_norm * 1.01 / radius)
+
+    outlines_dict["mask_pos"] = (mask_scale * head_x, mask_scale * head_y)
     clip_radius = radius * mask_scale
-    outlines_dict['clip_radius'] = (clip_radius,) * 2
-    outlines_dict['clip_origin'] = clip_origin
-    
+    outlines_dict["clip_radius"] = (clip_radius,) * 2
+    outlines_dict["clip_origin"] = clip_origin
+
     outlines = outlines_dict
-    
+
     return outlines
-
-
--- a/mne_icalabel/megnet/features.py
+++ b/mne_icalabel/megnet/features.py
@@ -1,29 +1,33 @@
-import numpy as np
-from scipy.spatial import ConvexHull
-from scipy import interpolate
-import matplotlib.pyplot as plt
 import io
-from PIL import Image
-from matplotlib.backends.backend_agg import FigureCanvasAgg as FigureCanvas
-import mne
-from mne.io import BaseRaw
-from mne.preprocessing import ICA
-from _utils import cart2sph, pol2cart
+
+import matplotlib.pyplot as plt
+import mne # type: ignore
+import numpy as np
+from mne.io import BaseRaw # type: ignore
+from mne.preprocessing import ICA # type: ignore
+from mne.utils import warn # type: ignore
 from numpy.typing import NDArray
-from mne.utils import warn
+from PIL import Image
+from scipy import interpolate # type: ignore
+from scipy.spatial import ConvexHull # type: ignore
+
+from _utils import cart2sph, pol2cart
 
 
-def get_megnet_features(raw:BaseRaw, ica:ICA):
+def get_megnet_features(raw: BaseRaw, ica: ICA):
     """
-    Extract time series and topomaps for each ICA component. the main work is focused on making BrainStorm-like topomaps which trained the MEGnet.
-    
+    Extract time series and topomaps for each ICA component.
+    the main work is focused on making BrainStorm-like topomaps
+    which trained the MEGnet.
+
     Parameters
     ----------
     raw : BaseRaw
-        The raw MEG data. The raw instance should have 250 Hz sampling frequency and more than 60 seconds.
+        The raw MEG data. The raw instance should have 250 Hz
+        sampling frequency and more than 60 seconds.
     ica : ICA
         The ICA object containing the independent components.
-    
+
     Returns
     -------
     time_series : np.ndarray
@@ -37,29 +41,32 @@ def get_megnet_features(raw:BaseRaw, ica:ICA):
             "Raw instance. The MEGnet model was fitted on"
             "MEG data and is not suited for other types of channels."
         )
-    
+
     if raw.times[-1] < 60:
         raise RuntimeError(
             f"The provided raw instance has {raw.times[-1]} seconds. "
-            "MEGnet was designed to classify features extracted from an MEG dataset "
-            "at least 60 seconds long. "
+            "MEGnet was designed to classify features extracted from "
+            "an MEG datasetat least 60 seconds long. "
         )
-    
-    if not np.isclose(raw.info['sfreq'], 250, atol=1e-1):
+
+    if not np.isclose(raw.info["sfreq"], 250, atol=1e-1):
         warn(
-            f"The provided raw instance is not sampled at 250 Hz (sfreq={raw.info['sfreq']} Hz). "
-            "MEGnet was designed to classify features extracted from an MEG dataset "
-            "sampled at 250 Hz (see the 'resample()' method for raw). "
+            "The provided raw instance is not sampled at 250 Hz"
+            f"(sfreq={raw.info['sfreq']} Hz). "
+            "MEGnet was designed to classify features extracted from"
+            "an MEG dataset sampled at 250 Hz"
+            "(see the 'resample()' method for raw)."
             "The classification performance might be negatively impacted."
         )
-        
+
     pos_new, outlines = _get_topomaps_data(ica)
     topomaps = _get_topomaps(ica, pos_new, outlines)
     time_series = ica.get_sources(raw)._data
 
     return time_series, topomaps
 
-def _make_head_outlines(sphere:NDArray, pos:NDArray, clip_origin:tuple):
+
+def _make_head_outlines(sphere: NDArray, pos: NDArray, clip_origin: tuple):
     """
     Generate head outlines and mask positions for the topomap plot.
     """
@@ -68,29 +75,35 @@ def _make_head_outlines(sphere:NDArray, pos:NDArray, clip_origin:tuple):
     head_x = np.cos(ll) * radius * 1.01 + x
     head_y = np.sin(ll) * radius * 1.01 + y
 
-    mask_scale = max(1., np.linalg.norm(pos, axis=1).max() * 1.01 / radius)
+    mask_scale = max(1.0, np.linalg.norm(pos, axis=1).max() * 1.01 / radius)
     clip_radius = radius * mask_scale
 
     outlines_dict = {
-        'head': (head_x, head_y),
-        'mask_pos': (mask_scale * head_x, mask_scale * head_y),
-        'clip_radius': (clip_radius,) * 2,
-        'clip_origin': clip_origin
+        "head": (head_x, head_y),
+        "mask_pos": (mask_scale * head_x, mask_scale * head_y),
+        "clip_radius": (clip_radius,) * 2,
+        "clip_origin": clip_origin,
     }
     return outlines_dict
 
-def _get_topomaps_data(ica:ICA):
+
+def _get_topomaps_data(ica: ICA):
     """
     Prepare 2D sensor positions and outlines for topomap plotting.
     """
-    mags = mne.pick_types(ica.info, meg='mag')
-    channel_info = ica.info['chs']
-    channel_locations_3d = np.array([channel_info[i]['loc'][0:3] for i in mags])
+    mags = mne.pick_types(ica.info, meg="mag")
+    channel_info = ica.info["chs"]
+    loc_3d = [channel_info[i]["loc"][0:3] for i in mags]
+    channel_locations_3d = np.array(loc_3d)
 
     # Convert to spherical and then to 2D
-    sph_coords = np.transpose(cart2sph(channel_locations_3d[:, 0], 
-                                       channel_locations_3d[:, 1], 
-                                       channel_locations_3d[:, 2]))
+    sph_coords = np.transpose(
+        cart2sph(
+            channel_locations_3d[:, 0],
+            channel_locations_3d[:, 1],
+            channel_locations_3d[:, 2],
+        )
+    )
     TH, PHI = sph_coords[:, 1], sph_coords[:, 2]
     newR = 1 - PHI / np.pi * 2
     channel_locations_2d = np.transpose(pol2cart(newR, TH))
@@ -100,7 +113,13 @@ def _get_topomaps_data(ica:ICA):
     border_indices = hull.vertices
     Dborder = 1 / newR[border_indices]
 
-    funcTh = np.hstack([TH[border_indices] - 2 * np.pi, TH[border_indices], TH[border_indices] + 2 * np.pi])
+    funcTh = np.hstack(
+        [
+            TH[border_indices] - 2 * np.pi,
+            TH[border_indices],
+            TH[border_indices] + 2 * np.pi,
+        ]
+    )
     funcD = np.hstack((Dborder, Dborder, Dborder))
     interp_func = interpolate.interp1d(funcTh, funcD)
     D = interp_func(TH)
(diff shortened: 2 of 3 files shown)


mscheltienne:
Thanks for the PR, it looks like a very good start!
If I understand correctly this model is provided within the BrainStorm toolbox?

Would it be possible to take a known input, load it in MNE, load it in BrainStorm; then run the pipeline on MNE (this PR) and on BrainStorm, export the output from BrainStorm and compare it to the output from MNE (this PR)?
This is the idea os a 'unit test', where the goal is to write a test which makes sure that the code does what we want. If you can write this comparison, we can easily turn it into a test in the `tests` folders.

colehank:
This [MEGnet](https://github.com/DeepLearningForPrecisionHealthLab/MegNET_2020) is an open-source model, independent of BrainStorm, built on a Keras-based architecture. The input includes the ICA topomap from BrainStorm, a 120x120 RGB array, along with the corresponding ICA time series array. The output provides the probabilities that the component is classified as brain/other, eye blink, eye movement, or heart.

I performed “Feature Simulation” because the model was trained on ICA topomaps RGB array from BrainStorm. When I used MNE ICA topomaps to generate RGB images, the classification performance was not good. To improve this, I ran simulations in features.py. Below is the result using MNE sample MEG data:
b796b475-ab78-4791-8e53-130367f892a2
These features demonstrated acceptable classification accuracy in the resulting classifications.

To fully follow the MEGnet pipeline for comparison, we could start with BrainStorm, going from raw ➡️ ICA ➡️ topomaps & time series, and then use the original Keras model alongside the method in this PR for comparison. However, I have never used BrainStorm before. Perhaps I can write a comparison using the feature extraction method from this PR (`features.py`) for both models and check whether the classification results are consistent between ONNX and Keras？

mscheltienne:
Thanks for the clarification. I'm not worried of a difference between Keras and ONNX, but indeed we can test this:

- Take a given input X
- Run it through both Keras and ONNX version
- Check for an exact match between the outputs

The good thing is that this is easy as the input X can be absolutely anything with the correct input shape and doesn't need to correspond to actual brain data.

-----------

Next, if this model was trained on topographic map from BrainStorm, I would expect it to behave differently if you feed slightly different topographic maps; as you comment. In this sense, we need to replicate the topographic map from BrainStorm, starting from an MNE object. This is what is done for the ICLabel model where those functions https://github.com/mne-tools/mne-icalabel/blob/main/mne_icalabel/features/topomap.py replicated the topographic map from EEGLab. To test the replication, we extracted topographic maps from EEGLab (in the end it's just a .mat file), and compared them with what we computed with the function in topomap.py.
It would be great to include the same kind of comparison between a topographic map outputted by brainstorm and one outputted by the feature code in your PR.

colehank:
Oh! I see, this will be a more systematic technical validation. I will follow these two steps to verify next. Thanks!

colehank:
Hi! I recently had some time to complete the two-step verification process. You can see the details in [a
temp repo](https://github.com/colehank/temp_mneicalabel/tree/main/megnet). The main result is shown in the figure below: the horizontal axis demonstrates the consistency between Keras and ONNX, while the vertical axis shows the generalizability of MEGnet across Brainstorm and MNE.

>Additionally, I changed the original method for plotting the topomap by removing the contours, which seems to have improved MEGnet’s recognition accuracy. This is likely because the BrainStorm topomaps used to train MEGnet had very thin contour lines, which were white and almost invisible (as shown in the figure below).

2ea020c8-ae9a-4d6c-93b6-6ebd420a29be

colehank:
Apologies for the multiple pushes😢. This is my first time creating a PR, and I’m still not very familiar with Git.

adam2392:
Thanks for this work @colehank!

If I understand the code in your temp repo: Presumably the main difference is in the computation of the ICA, and corresponding topomaps within BrainStorm GUI?

Is it possible to also compare the MEGNet within BrainStorm and the MEGNet within mne-icalabel (this PR)? output on the same topomaps? E.g. using either/or the topomaps you computed from BrainStorm, or the topomaps from mne-icalabel?

This would verify that the network is fully the same. One thing you can do is save 1 input topomap, and 1 output from the network, and we can just use this "small test data" as a unit-test within this repo.  For example, we do this [here](https://github.com/mne-tools/mne-icalabel/blob/3a14b3400f97c4665d1afe4e9d81808ca09ad740/mne_icalabel/iclabel/network/tests/test_network.py#L134-L175) for ICLabel.

Now, the next, more difficult question is how to proceed if the topomaps are different. My question would be: i) are there substantial differences that will lead to performance changes between MNE approach vs BrainStorm approach? and ii) if not, do we care about matching exactly that in BrainStorm.

mscheltienne:
> Apologies for the multiple pushes😢. This is my first time creating a PR, and I’m still not very familiar with Git.

Don't worry about it :)

I started a thorough review on Thursday, before I got interrupted by faults on our MEG system 😢 
I'll try to complete it this week.

@adam2392 That's one part I did not get at the beginning either, but the model/network is *not* available in BrainStorm. The people who developed it made a strange choice: create topographic map in brainstorm, export them as PNG of shape (120, 120, 3) and export the ICA traces; then train a network in Python on the colored images and on the traces. Thus, the network is already in Python, not much to compare here ;)

mscheltienne (commented):
I'll post those initial comments from last Thursday, you could already start tackling them :)

mne_icalabel/megnet/features.py
```diff
@@ -0,0 +1,169 @@
...
+        )
+
+    if raw.times[-1] < 60:
+        raise RuntimeError(
+            f"The provided raw instance has {raw.times[-1]} seconds. "
+            "MEGnet was designed to classify features extracted from "
+            "an MEG datasetat least 60 seconds long. "
+        )
+
+    if not np.isclose(raw.info["sfreq"], 250, atol=1e-1):
```
colehank, line 53: Would it be better to use warn instead of raise in the presence of linenoise? Based on my experience, linenoise does not affect ICA's ability to handle eye movements, saccades, and heartbeat ICs, so it should not impact MEGnet's performance either, as MEGnet is only sensitive to these artifacts.

mne_icalabel/megnet/features.py
```diff
@@ -0,0 +1,169 @@
...
+        warn(
+            "The provided raw instance is not sampled at 250 Hz"
+            f"(sfreq={raw.info['sfreq']} Hz). "
+            "MEGnet was designed to classify features extracted from"
+            "an MEG dataset sampled at 250 Hz"
+            "(see the 'resample()' method for raw)."
+            "The classification performance might be negatively impacted."
+        )
```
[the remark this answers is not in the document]
colehank: I think I’ve solved these issues! Specifically, for detecting linenoise, I wrote a function called _check_notch, and I tested it with good results.

mne_icalabel/megnet/features.py
```diff
@@ -0,0 +1,169 @@
...
+
```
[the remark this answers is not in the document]
colehank: For MEGnet, they used infomax, so I only implemented detection specifically for the infomax method here.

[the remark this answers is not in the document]
colehank: Thankx! For MEGnet, they used infomax, so I only implemented detection specifically for the infomax method here.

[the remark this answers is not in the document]
colehank: Although MEGnet was trained on 20 ICs for each raw data segment, the model makes judgments based on a single topomap and timeseries. Do we need to raise or warn if the ICA does not have 20 components?

mne_icalabel/megnet/features.py
```diff
@@ -0,0 +1,169 @@
...
+        )
+        img_buf = io.BytesIO()
+        mnefig.figure.savefig(
+            img_buf, format="png", dpi=120, bbox_inches="tight", pad_inches=0
+        )
+        img_buf.seek(0)
+        rgba_image = Image.open(img_buf)
+        rgb_image = rgba_image.convert("RGB")
+        img_buf.close()
+        plt.close(fig)
```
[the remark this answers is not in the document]
colehank: Yes, I'm also puzzled as to why RGB colors were used instead of the original channel ICA weights to represent the topomap. However, it's good that the model's performance is still within an acceptable range.

mne_icalabel/megnet/label_componets.py
```diff
@@ -0,0 +1,120 @@
...
+                ).astype(np.float32),
+            }
+            prediction = session.run(None, onnx_inputs)[0]
+            weighted_predictions[start_time] = prediction * chunk_votes[start_time]
+
+        comp_prediction = np.stack(list(weighted_predictions.values())).mean(axis=0)
+        comp_prediction /= comp_prediction.sum()
+        predction_vote.append(comp_prediction)
+
+    return np.stack(predction_vote)
```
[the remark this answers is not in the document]
colehank: Yes, I did use [MEGnet's chunk predicting algorithm](https://github.com/DeepLearningForPrecisionHealthLab/MegNET_2020/blob/main/MEGnet/megnet_utilities.py) here, along with some simplifications. The original function also included model evaluation, among other things.

colehank:
Thanks for your suggestions! @mscheltienne @adam2392 

There are indeed differences in the implementation of infomax between BrainStorm and MNE when we plot components, but from my experience, this difference doesn't seem to affect MEGnet's classification performance, as demonstrated in the fig before.

Based on @mscheltienne  comments, I've made an updated iteration, and I hope it meets the required functionalities.

mne_icalabel/megnet/features.py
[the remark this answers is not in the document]
colehank: Oh, I read

> Thus, the choice of 20 ICA components presents a reasonable balance: it provides enough components to account for both the variability due to canonical biological networks and artifact sources while not dividing components into unrecognizable waveforms. In this study, the predictive models use both the spatial map and the time course components as inputs.

Maybe raising is a reasonable decision; I previously missed the description about this in the article.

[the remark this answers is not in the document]
mscheltienne: Actually, both of your replies have a good argument. I think a warning with a very light message won't hurt anyway.

mscheltienne:
Thanks for tackling my comments, I'll try to continue the review as soon as possible!

mne_icalabel/megnet/features.py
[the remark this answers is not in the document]
colehank: Alright then, I’ll include a warning in the next version~

mne_icalabel/megnet/features.py
```diff
@@ -0,0 +1,291 @@
...
+    threshold_factor : float, optional
+        Multiplicative factor for setting the detection threshold.
+        The threshold is set as the mean of the neighboring frequencies plus
+        `threshold_factor` times the standard deviation. Default is 3.
+    show : bool, optional
+        Whether to plot the PSD for channels affected by line noise.
+
+    Returns
+    -------
+    bool
```
colehank, line 227: I made a mistake here, this function returns a `list` contains channels with line noise

mne_icalabel/megnet/_utils.py
```diff
@@ -0,0 +1,44 @@
...
+    r = np.sqrt(x * x + y * y + z * z)
+    theta = np.arctan2(y, x)
+    phi = np.arctan2(z, xy)
+    return r, theta, phi
+
+
+def pol2cart(rho, phi):
+    x = rho * np.cos(phi)
+    y = rho * np.sin(phi)
+    return x, y
```
adam2392, line 17: These functions are also defined for ICLabel, so I wonder if we can pull them out for a general utility functions related to geometry.

https://github.com/mne-tools/mne-icalabel/blob/b2fc448dd2a5718eeeefee14d051525f2cb4b209/mne_icalabel/iclabel/_utils.py#L97

mne_icalabel/megnet/_utils.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+def make_head_outlines(sphere, pos, outlines, clip_origin):
```
adam2392, line 20: Can you write a docstring for the function?

mne_icalabel/megnet/features.py
```diff
@@ -0,0 +1,291 @@
...
+    ----------
+    raw : BaseRaw
+        The raw MEG data. The raw instance should have 250 Hz
+        sampling frequency and more than 60 seconds.
+    ica : ICA
+        The ICA object containing the independent components.
+
+    Returns
+    -------
+    time_series : np.ndarray
```
adam2392, line 33:
```suggestion
    time_series : np.ndarray of shape (n_components, n_times)
```

Is this right?

mne_icalabel/megnet/features.py
```diff
@@ -0,0 +1,291 @@
...
+        The time series for each ICA component.
+    topomaps : np.ndarray
```
adam2392, line 35:
```suggestion
    topomaps : np.ndarray of shape (n_components, width, height, channels)
```

Is this right?

mne_icalabel/megnet/features.py
```diff
@@ -0,0 +1,291 @@
...
+        plt.ylabel("Power Spectral Density (PSD)")
+        plt.legend()
+        plt.show()
+
+    return line_noise_detected
+
+
+def _check_notch(
+    raw: BaseRaw, picks: str, neighbor_width: float = 2.0, threshold_factor: float = 3
+) -> bool:
```
adam2392, line 284: I'm wondering why we don't just notch filter for them at the `line_noise` and its harmonics?

Would this impact the results?

I guess I can imagine a user constructing the Raw object via `RawArray`, which wouldn't need them to specify the line noise frequency, so maybe this wouldn't work all the time.

mne_icalabel/megnet/label_components.py
```diff
@@ -0,0 +1,120 @@
...
+from mne.preprocessing import ICA
+from numpy.typing import NDArray
+
+from .features import get_megnet_features
+
+
+def megnet_label_components(
+    raw: BaseRaw,
+    ica: ICA,
+    model_path: str = op.join("assets", "network", "megnet.onnx"),
```
adam2392, line 15: Do we need to expose the model path? I presume the model is always going to use this model?

Perhaps the API can look similar to ICLabel, and add the `inplace` argument?

https://github.com/mne-tools/mne-icalabel/blob/b2fc448dd2a5718eeeefee14d051525f2cb4b209/mne_icalabel/iclabel/label_components.py#L12

adam2392 (commented):
Left a few comments. Overall looking good and thanks for answering questions!

mne_icalabel/megnet/features.py
mscheltienne: Elekta/MEGIN system do store the line noise frequency, but I don't know if this is the case for other systems. I would keep it is as it is now, with the responsibility to apply the notch filter left to the user.

mne_icalabel/megnet/_utils.py
mscheltienne: `pol2cart` is indeed a duplicate, `cart2sph` returns the element in a different order and it's a bit annoying to change the order. We can keep code de-duplication for a future PR.

colehank:
Thank you all for your helpful suggestions, and apologies for the delay in my response as I’ve been resting due to illness.

I’ve just submitted the unit test script for the main functions in `label_components` and `features`.

Additionally, I made the following updates:

- The `megnet_label_components` now outputs an array with shape `(n_components, n_classes)`. However, please note that the classes for MEGNET are `['brain/other', 'eye movement', 'heart beat', 'eye blink']`, which do not exactly match the classes used in ICLabel.

- I added a docstring for the `_make_head_outlines` function.

- The `pol2cart` function has been updated to use `_pol2cart` from `mne_icalabel.iclabel._utils`.

colehank:
first time writing `pytest`, I might made some silly mistakes, but I did my best to ensure that there are no failures in the `megnet` directory locally. However, when I submitted it to `pre-commit.ci`, I encountered the following error:

`_MODEL_PATH: str = files("mne_icalabel.megnet") / "assets" / "megnet.onnx" `
`TypeError: expected str, bytes or os.PathLike object, not NoneType`

I'm not quite sure why this is happening... It seems that the model path is not being recognized correctly. Could it be because my branch is not in sync with the main branch?

mscheltienne:
Thanks, I'll try to have a look tomorrow!

> the classes for MEGNET are ['brain/other', 'eye movement', 'heart beat', 'eye blink'], which do not exactly match the classes used in ICLabel.

That's fine and expected. 

Looking briefly at the test I see:
- Failure in `pre-commit.ci`:

```
codespell................................................................Failed
- hook id: codespell
- exit code: 65

mne_icalabel/megnet/features.py:34: hould ==> hold, should
```

You'll find line 34 in this file a spelling error, suggestions are `hold` or `should`.
If this is not a spelling error, we can tell `codespell` to ignore it.

- Failure in `pytest`:

```
     def test_megnet_label_components(raw_ica):
        """Test whether the function returns the correct artifact index."""
        real_atrifact_idx = [0, 3, 5]  # heart beat, eye movement, heart beat
        prob = megnet_label_components(*raw_ica)
        this_atrifact_idx = list(np.nonzero(prob.argmax(axis=1))[0])
>       assert this_atrifact_idx == real_atrifact_idx
E       assert [np.int64(0),..., np.int64(5)] == [0, 3, 5]
E         
E         At index 2 diff: np.int64(4) != 5
E         Left contains one more item: np.int64(5)
E         
E         Full diff:
E           [
E         -     0,
E         -     3,
E         -     5,
E         +     np.int64(0),
E         +     np.int64(3),
E         +     np.int64(4),
E         +     np.int64(5),
E           ]

mne_icalabel/megnet/tests/test_label_components.py:38: AssertionError
```

mscheltienne:
No need to apologizes, that's fine and it does seem to be environment issues. The online CIs don't suffer from those, so they will tell us if `pytest` is passing or not. I just-restarted the workflows, let's see!

mne_icalabel/megnet/features.py
```diff
@@ -0,0 +1,217 @@
...
+
+def get_megnet_features(raw: BaseRaw, ica: ICA):
+    """Extract time series and topomaps for each ICA component.
+
+    MEGNet uses topomaps from BrainStorm exported as 120x120x3 RGB images.
+    Thus, we need to replicate the 'appearance'/'look' of a BrainStorm topomap.
+
+    Parameters
+    ----------
+    raw : Raw.
```
adam2392, line 27:
```suggestion
    raw : Raw
```

mne_icalabel/megnet/label_components.py
```diff
@@ -0,0 +1,107 @@
...
+
+import numpy as np
+import onnxruntime as ort
+from mne.io import BaseRaw
+from mne.preprocessing import ICA
+from numpy.typing import NDArray
+
+from .features import get_megnet_features
+
+_MODEL_PATH: str = files("mne_icalabel.megnet") / "assets" / "megnet.onnx"
```
adam2392, line 11: I think this line causes issues. 

See: https://github.com/mne-tools/mne-icalabel/blob/43eec6dbce9cf9267e1f2de96d0adc3835c92268/mne_icalabel/iclabel/network/onnx.py#L27

Let us know if that doesn't work.

adam2392 (commented):
I think it's looking good. Just have to wrangle with our testing framework. 

Thanks for the hard efforts @colehank!

mne_icalabel/megnet/label_components.py
mscheltienne: I don't think he modified anything from `iclabel`, that should not be failing.

adam2392: Yes nvm. I misread the error. 

Hmm it seems the file is somehow not found.

colehank: Oh, I found the issue! After I added the `__init__.py `file in the `mne_icalabel/megnet` directory, it was able to correctly find `megnet.onnx`. (but I remember that python(at least v3.8)  don't require __init__.py to treat a folder as a package.)

mne_icalabel/megnet/tests/test_label_components.py
```diff
@@ -0,0 +1,67 @@
...
+    ica.fit(raw)
+
+    return raw, ica
+
+
+def test_megnet_label_components(raw_ica):
+    """Test whether the function returns the correct artifact index."""
+    real_atrifact_idx = [0, 3, 5]  # heart beat, eye movement, heart beat
+    prob = megnet_label_components(*raw_ica)
+    this_atrifact_idx = list(np.nonzero(prob.argmax(axis=1))[0])
```
colehank, line 37: It's a bit strange here. When I ran it yesterday, the model detection results were 0, 3, 5; but today they have become 0, 3, 4, 5, which caused the pytest to fail. It might be due to the randomness in the ICA algorithm？ (even though I set the random_state).

colehank: Maybe we can try asserting `set(real_atrifact_idx).issubset(set(this_atrifact_idx))` to ensure that the true labels are a subset of the model's predicted labels. This is a debugging-specific approach, but it might not be entirely reasonable...

mscheltienne: If your ICA has a fix random state, and if the model is deterministic (which I think it is?), you should get the exact same results.

To debug you can try:
- run 2/3 times the 'icai' part and compare the mixing/unmixing matrixes with `assert_allclose`. They should be identical.
- run 2/3 times the model part and compare the model outputs (not the index, but the probabilities per class), they should be identical.

mscheltienne: It's a floating point rounding error:

```
In [12]: prob[4, 0]
Out[12]: np.float64(0.4999999999999503)

In [13]: prob[4, 1]
Out[13]: np.float64(0.5000000000000497)
```

mscheltienne:
@colehank @adam2392 Almost a year after it was open, I had some times today to push it over the finnish line. The changes were mostly:

- Documentation, adding the proper RST pages, bilbliography entries, etc..
- Finding that the test faling was due to a floating point error (see: https://github.com/mne-tools/mne-icalabel/pull/207#discussion_r1844312276)
- Moving the coordinate transformation utilities to a single `utils/transform.py` file

I'll let it open for a bit, in case you want to check, and then I'll merge and release `0.8.0`.

colehank (commented):
Looks great! Many thanks for the reviews and feedback — happy I could contribute to mne-icalabel.

PR #316, opened 2026-07-11
fix: don't warn about missing CAR when average reference is a projection

Leonard013:
Fixes #290.

**Problem.** `get_iclabel_features` (and therefore `label_components`) warns that the data "does not seem to be referenced to a common average reference (CAR)" based on `inst.info["custom_ref_applied"] == 0`. But that flag is only set when an average reference is applied *directly* (`set_eeg_reference("average", projection=False)`). When the CAR is instead added as an SSP **projection** (`set_eeg_reference("average", projection=True)`, whether left pending or later applied with `apply_proj()`), `custom_ref_applied` stays `0`, so the warning fired even though a common average reference had been requested/applied — the false warning reported in #290.

**Fix.** Also treat the presence of an average-reference projector (pending or applied) as a CAR, using mne's `_has_eeg_average_ref_proj` (its default `check_active=False` detects both states). This matches @mscheltienne's diagnosis in the issue. The import mirrors the existing `check_version("mne", "1.6")` gate already used in `mne_icalabel/features/topomap.py` for the same `mne._fiff` vs `mne.io` reorg boundary (the helper has existed since mne 1.2, only the module path changed at 1.6), so the declared `mne>=1.2` floor is preserved.

Out of scope (unchanged, noted for transparency): the `custom_ref_applied == 0` check also treats the CSD state (`FIFFV_MNE_CUSTOM_REF_CSD == 2`) as "don't warn", and the broader limitation that the reference isn't fully stored in `info` (mne-python#8962, referenced by the existing TODO) remains.

**Regression test** (`test_features.py::test_get_iclabel_features_car_warning_reference_scenarios`): a small synthetic `RawArray` (standard_1020 montage, no dataset download, no torch/onnx) with a fitted extended-infomax ICA, asserting the CAR warning across four reference scenarios — no reference (warns, guarding existing behavior), direct average (no warn), pending average projection (must not warn — the bug), applied average projection (must not warn — the bug). It fails on `main` (scenarios 3 & 4) and passes with this change.

Verified locally on macOS arm64 (Python 3.12, mne 1.12.1): the new test fails before / passes after; the full `test_features.py` passes (22 tests); `ruff check` and `ruff format --check` are clean.

Prepared with AI assistance (Claude Code); the change and its verification are described in full above.

### DIFF
--- a/doc/changes/authors.inc
+++ b/doc/changes/authors.inc
@@ -4,3 +4,4 @@
 .. _Anand Saini: https://github.com/anandsaini024
 .. _Scott Huberty: https://github.com/scott-huberty
 .. _Guohao Zhang: https://github.com/colehank
+.. _Leonardo Scappatura: https://github.com/Leonard013
--- a/doc/changes/latest.rst
+++ b/doc/changes/latest.rst
@@ -16,3 +16,5 @@
 
 Version 0.10
 ============
+
+- Fix a spurious warning from :func:`~mne_icalabel.iclabel.get_iclabel_features` (and :func:`~mne_icalabel.label_components`) about the data not being referenced to a common average reference (CAR) when the average reference was applied as a projection (``set_eeg_reference("average", projection=True)``) rather than directly (:pr:`316` by `Leonardo Scappatura`_)
--- a/mne_icalabel/iclabel/features.py
+++ b/mne_icalabel/iclabel/features.py
@@ -5,8 +5,13 @@ from typing import TYPE_CHECKING
 import numpy as np
 from mne import BaseEpochs
 from mne.io import BaseRaw
-from mne.utils import warn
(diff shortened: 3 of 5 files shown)


mscheltienne:
Pushed a small commit moving the imports to the module-level. As I said in https://github.com/mne-tools/mne-icalabel/issues/290, the question is how is this actually used further down-the-line since the intent is for ICLabel to receive common average referenced data.

With an un-applied projector, it will allow non CAR-data to flow into ICLabel, `get_iclabel_features` consume it to compute the activations:

```
data = inst.get_data(picks=ica.ch_names) * 1e6
icaact = weights @ data
```

So, in the case of this PR, we need to set `check_active=True`. Pushed as a second commit.

